PR #6, opened 2023-05-30
Added preferred maximum-path configuration from broadcom

Cellebyte:
* Add suggestions from broadcom via Environment Variable.
* hotfix.

### DIFF
--- a/pkg/frr/configure.go
+++ b/pkg/frr/configure.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"io/ioutil"
 	"os"
+	"strconv"
 
 	"github.com/telekom/das-schiff-network-operator/pkg/nl"
 )
@@ -24,6 +25,7 @@ type FRRTemplateConfig struct {
 	Hostname         string
 	UnderlayRouterID string
 	HostRouterID     string
+	ECMPMaximumPaths int
 }
 
 func (m *FRRManager) Configure(in FRRConfiguration) (bool, error) {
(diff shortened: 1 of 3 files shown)


pkg/frr/configure.go
```diff
@@ -91,15 +93,26 @@ func (f *FRRManager) renderSubtemplates(in FRRConfiguration) (*FRRTemplateConfig
 	if err != nil {
 		return nil, err
 	}
+	envECMPMaximumPaths, isSet := os.LookupEnv("NWOP_ECMP_MAXIMUM_PATHS")
```
schrej, line 96: FYI: The second return value is commonly called `ok` for things like this, but no need to change that.

pkg/frr/configure.go
```diff
@@ -91,15 +93,26 @@ func (f *FRRManager) renderSubtemplates(in FRRConfiguration) (*FRRTemplateConfig
...
+	ecmpMaximumPaths := 8 // Default to BCMs preferred target of maximum 8 paths per route.
+	if isSet {
+		parsedECMPMaximumPaths, err := strconv.Atoi(envECMPMaximumPaths)
+		if err != nil {
+			ecmpMaximumPaths = parsedECMPMaximumPaths
+		} else {
+			return nil, err
+		}
+	}
```
schrej, line 105: I would implement it like this, which is a more common style and reduces scope of some of the variables.

In addition the if condition for error handling was the wrong way round, which is corrected here as well.

```suggestion
	ecmpMaximumPaths := 8 // Default to BCMs preferred target of maximum 8 paths per route.
	if envECMPMaximumPaths, ok := os.LookupEnv("NWOP_ECMP_MAXIMUM_PATHS"); ok {
		parsedECMPMaximumPaths, err := strconv.Atoi(envECMPMaximumPaths)
		if err != nil {
			return nil, err
		}
		ecmpMaximumPaths = parsedECMPMaximumPaths
	}
```

schrej (changes requested):
One bug, lgtm otherwise

schrej (approved):
go part lgtm

Cellebyte:
should we still add this feature as a global configuration or not?

MaxRink:
@Cellebyte we should default to something sane

Cellebyte:
> @Cellebyte we should default to something sane

What is something sane @MaxRink ?

Cellebyte:
Closed as not needed

PR #23, opened 2023-06-19
Introduce periodic requeue

chdxD1:
Requeue the VRF (L3VNI) and L2VNI reconciles after 10 minutes. This will run a sync with netlink every 10 minutes


schrej (commented):
We could also use `SyncPeriod` which is part of Manager's options to cause everyting to be reconciled every 10 minutes. Not sure what's better here.

schrej (approved):
Should've read further.

PR #27, opened 2023-06-19
Introduce Central Reconciler

chdxD1:
Previously reconciliation happened in each controller. However we have an interdependency between L2VNIs and L3VNIs. Thus we summarize reconciliation in a central place. This also allows us to reference L2VNIs in FRR config (e.g. for RT settings). Also we now reconcile them after L3VNI reconciliation, attaching VRFs if needed.

### DIFF
--- a/main.go
+++ b/main.go
@@ -38,6 +38,7 @@ import (
 	"github.com/telekom/das-schiff-network-operator/pkg/config"
 	"github.com/telekom/das-schiff-network-operator/pkg/macvlan"
 	"github.com/telekom/das-schiff-network-operator/pkg/notrack"
+	"github.com/telekom/das-schiff-network-operator/pkg/reconciler"
 	//+kubebuilder:scaffold:imports
 )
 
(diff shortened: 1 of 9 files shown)


main.go
```diff
@@ -97,25 +98,30 @@ func main() {
 	anycastTracker := &anycast.AnycastTracker{}
 
 	// Start VRFRouteConfigurationReconciler when we are not running in only BPF mode.
```
schrej, line 100: comment should probably move down a bit

pkg/reconciler/layer2.go
```diff
@@ -0,0 +1,157 @@
...
+	for _, item := range layer2List.Items {
+		if item.Spec.NodeSelector != nil {
+			selector, err := metav1.LabelSelectorAsSelector(item.Spec.NodeSelector)
+			if err != nil {
+				r.Logger.Error(err, "error converting nodeSelector of layer2 to selector", "layer2", item.ObjectMeta.Name)
+				return nil, err
+			}
+			if !selector.Matches(labels.Set(node.ObjectMeta.Labels)) {
+				r.Logger.Info("local node does not match nodeSelector of layer2", "layer2", item.ObjectMeta.Name, "node", nodeName)
+				return nil, err
```
schrej, line 44: I think this is supposed to be a `continue`

pkg/reconciler/layer2.go
```diff
@@ -0,0 +1,157 @@
...
+		}
+
+		l2vnis = append(l2vnis, item)
+	}
+
+	return l2vnis, nil
+}
+
+func (r *reconcile) reconcileLayer2(l2vnis []networkv1alpha1.Layer2NetworkConfiguration) error {
+	layer2Info := []nl.Layer2Information{}
```
schrej, line 55: maybe call this `desired`, then it's a little clearer further down when comparing `existing` vs `desired`

pkg/reconciler/layer3.go
```diff
@@ -0,0 +1,214 @@
...
+
+	for _, vrf := range l3vnis {
+		spec := vrf.Spec
+
+		var vni int
+		var rt string
+
+		if val, ok := r.config.VRFToVNI[spec.VRF]; ok {
+			vni = val
+			r.Logger.Info("Configuring VRF from old VRFToVNI", "vrf", spec.VRF, "vni", val)
```
schrej, line 38: old has precedence over new? sounds wrong.

pkg/reconciler/layer3.go
```diff
@@ -0,0 +1,214 @@
...
+		config := vrfConfigMap[spec.VRF]
+
+		if len(spec.Export) > 0 {
+			config.Export = append(config.Export, handlePrefixItemList(spec.Export, spec.Seq))
+		}
+		if len(spec.Import) > 0 {
+			config.Import = append(config.Import, handlePrefixItemList(spec.Import, spec.Seq))
+		}
+		for _, aggregate := range spec.Aggregate {
+			_, network, _ := net.ParseCIDR(aggregate)
```
schrej, line 69: the CIDRs are always correct? Is that validated?

pkg/reconciler/layer3.go
```diff
@@ -0,0 +1,214 @@
...
+		return nil, err
+	}
+
+	// Check for VRFs that are configured on the host but no longer in Kubernetes
+	delete := []nl.VRFInformation{}
+	for _, cfg := range existing {
+		stillExists := false
+		for _, vrf := range vrfConfigs {
+			if vrf.Name == cfg.Name && vrf.VNI == cfg.VNI {
+				stillExists = true
```
schrej, line 140: you can probably add a break afterwards

pkg/reconciler/layer3.go
```diff
@@ -0,0 +1,214 @@
...
+	create := []nl.VRFInformation{}
+	for _, vrf := range vrfConfigs {
+		// Skip VRF with VNI SKIP_VRF_TEMPLATE_VNI
+		if vrf.VNI == config.SKIP_VRF_TEMPLATE_VNI {
+			continue
+		}
+		alreadyExists := false
+		for _, cfg := range existing {
+			if vrf.Name == cfg.Name && vrf.VNI == cfg.VNI {
+				alreadyExists = true
```
schrej, line 158: break as well

pkg/reconciler/layer3.go
```diff
@@ -0,0 +1,214 @@
...
+	prefixList := frr.PrefixList{
+		Seq: seq + 1,
+	}
+	for i, item := range input {
+		prefixList.Items = append(prefixList.Items, copyPrefixItemToFRRItem(i, item))
+	}
+	return prefixList
+}
+
+func copyPrefixItemToFRRItem(i int, item networkv1alpha1.VrfRouteConfigurationPrefixItem) frr.PrefixedRouteItem {
```
schrej, line 199: would call it `n` instead of `i`, since `i` is often used in loops, and `n` for arbitrary numbers or counts.

schrej (changes requested):

schrej (approved):

PR #28, opened 2023-07-03
Re-execute debounce when called during execution

chdxD1:


pkg/debounce/debounce.go
```diff
@@ -11,7 +11,8 @@ import (
 // Debouncer struct
 type Debouncer struct {
 	// Used for atomic operations
-	scheduled int32
+	scheduled             *atomic.Bool
+	calledDuringExecution *atomic.Bool
```
schrej, line 15: Why are these pointers? I think these can be regular values. They don't require initialisation that way as well.

pkg/debounce/debounce.go
```diff
@@ -34,8 +36,12 @@ func (d *Debouncer) debounceRoutine(ctx context.Context) {
 
 		err := d.function(ctx)
 		if err == nil {
-			// Reset scheduled to 0
-			atomic.StoreInt32(&d.scheduled, 0)
+			if d.calledDuringExecution.CompareAndSwap(true, false) {
+				d.debounceRoutine(ctx)
+			} else {
+				// Reset scheduled to 0
```
schrej, line 42: leftover I guess

chdxD1: yeah, better comments now because the use of atomic might not be intuitive in the beginning

schrej (approved):

Issue #32, opened 2023-07-07
Network healthchecks


MaxRink:
Ideally we dont have applications schedule on nodes that have broken networking.

based upon #31 network-operator could perform basic checks like "are all interfaces that i expect up and running"
"Is FRR running" or "can i reach x" before removing the taint from the node

Failing those checks would ideally emit both events and metrics for why its failing

chdxD1:
- Are all Interfaces up and running:
  We should be able to pass a list of interfaces to check. We are most interested in hardware interfaces
- Is FRR running
  a check using systemd is fine (systemd is already mounted because it is used for reload anyway)
- Can I reach X
  An address + port can be passed to the operator to check basic TCP reachability (which should suffice for our case)

Issue #31, opened 2023-07-07
Implement CPI Taints


MaxRink:
Ideally not Applications are scheduled until network-operator has run successfully
This can be done by utilizing the CPI tains
`node.cloudprovider.kubernetes.io/uninitialized`

This taint needs to be removed after the first successful run

PR #35, opened 2023-08-09
Implement network healthcheck

### LINKED ISSUE #41: [Bug] Discover the node via its HOSTNAME is not very practical if the hostname differs from the k8s nodename.
## Description

Error from the operator log.
```log
github.com/telekom/das-schiff-network-operator/pkg/debounce.(*Debouncer).debounceRoutine
	/workspace/pkg/debounce/debounce.go:35
2023-08-28T09:27:48.495Z	ERROR	controller.vrfrouteconfiguration	error debouncing	{"reconciler group": "network.schiff.telekom.de", "reconciler kind": "VRFRouteConfiguration", "name": "bla", "namespace": "", "error": "Node \"customer-1-md-md-0-bb44bbb9fxzlnbj-lvpd9.site.location.example.com\" not found"}
```
```console
$ kubectl get nodes

NAME                                   STATUS   ROLES           AGE     VERSION
customer-1-7jnmh                           Ready    control-plane   2d20h   v1.25.12
customer-1-md-md-0-bb44bbb9fxzlnbj-6vkrl   Ready    worker          2d20h   v1.25.12
customer-1-md-md-0-bb44bbb9fxzlnbj-dxc5f   Ready    worker          2d19h   v1.25.12
customer-1-md-md-0-bb44bbb9fxzlnbj-lvpd9   Ready    worker          2d20h   v1.25.12
customer-1-w7k6b                           Ready    control-plane   2d20h   v1.25.12
customer-1-wk5dt                           Ready    control-plane   2d20h   v1.25.12

# login on the node
root@customer-1-md-md-0-bb44bbb9fxzlnbj-lvpd9:/# hostname
customer-1-md-md-0-bb44bbb9fxzlnbj-lvpd9.site.location.example.com
root@customer-1-md-md-0-bb44bbb9fxzlnbj-lvpd9:/# hostname --fqdn
customer-1-md-md-0-bb44bbb9fxzlnbj-lvpd9.site.location.example.com

root@customer-1-md-md-0-bb44bbb9fxzlnbj-lvpd9:/# hostnamectl status
   Static hostname: customer-1-md-md-0-bb44bbb9fxzlnbj-lvpd9.site.location.example.com
   Pretty hostname: localhost.local
         Icon name: computer-server
           Chassis: server
        Machine ID: cd9f27f0e96c4d74a3e4f37404cbc802
           Boot ID: b6139e6f644c4fa4a7a21d41cdbcfa22
  Operating System: Ubuntu 20.04.6 LTS
            Kernel: Linux 5.15.0-79-generic
      Architecture: x86-64

```

## Findings

Currently the operator discovers the node it is running on via the `HOSTNAME` environment variable.
This becomes problematic, when the hostname differs from the nodename in kubernetes.

Calico is using a spec field ref into their environment Variables called `NODENAME`.

## Suggestion
Use something like this in the deployment of network operators daemonset to ensure the environment variable is always correctly set for reconciliation loops.

```yaml
        - name: NODENAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName
```

p-strusiewiczsurmacki-mobica:
This PR implements networking healthcheck as described in #31 and #32.
It also fixes #41.

CPI taint will be removed once all networking checks pass. This is done after first reconciliation loop iteration for 'full' pods and at the end of main() function for bpf-only pods.

Implemented checks are:

- FRR - checks if FRR daemon is active and in running state (will not be used on bpf-only pods)
- Interfaces - checks if provided interfaces are in UP state
- Reachability - checks if TCP connection can be established to provided hosts on selected ports (note: if connection is refused, but host is reachable then test will still pass). Connection timeout can be configured.

All data can be provided using _optional_ config file. Config file location is by default `/opt/network-operator/net-healthcheck-config.yaml` but can be changed with env variable `OPERATOR_NETHEALTHCHECK_CONFIG`. However changing path will make the config file mandatory.

Additionally unit tests were created for the new `healthcheck` package.


pkg/healthcheck/healthcheck_test.go
```diff
@@ -0,0 +1,355 @@
...
+	RunSpecsWithDefaultAndCustomReporters(t,
+		"HealthCheck Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = Describe("HealthCheck", func() {
+	var _ = Context("LoadConfig() should", func() {
+		var _ = It("return error if config is mandatory but does not exist", func() {
+			t := testing.T{}
+			t.Setenv(CONFIG_ENV, "/some/invalid/path")
```
schrej: I think constructing your own `testing.T` doesn't work. Just use `Expect(os.Setenv()).To(Succeed())` here instead

pkg/healthcheck/healthcheck_test.go
```diff
@@ -0,0 +1,355 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			_, err = filePath.WriteString(fakeConfInvalid)
+			Expect(err).ToNot(HaveOccurred())
+			_, err = LoadConfig(filePath.Name())
+			Expect(err).To(HaveOccurred())
+		})
+		var _ = It("return no error if valid config exists", func() {
+			filePath, err := os.CreateTemp(tmpPath, "healtchecktest")
+			Expect(err).ToNot(HaveOccurred())
+			_, err = filePath.WriteString(fakeConf)
```
schrej, line 68: for the fakeconf you could also just create a `testdata` directory in this package, put it in there and commit to git, and then load using a relative path

pkg/healthcheck/healthcheck_test.go
```diff
@@ -0,0 +1,355 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(hc).ToNot(BeNil())
+			Expect(hc.IsInitialized()).To(BeFalse())
+			err = hc.CheckReachability()
+			Expect(err).ToNot(HaveOccurred())
+		})
+	})
+})
+
+type fakeFRRManagerError struct{}
```
schrej, line 269: instead of writing mocks yourself, you could also have a look at https://github.com/uber-go/mock

It's pretty nice, and allows to register handlers to return custom values for testing and so on, so you only need to generate one mock and can use it for everything. Might be a bit overkill here though.

pkg/healthcheck/healthcheck_test.go
```diff
@@ -0,0 +1,355 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(hc).ToNot(BeNil())
+			Expect(hc.IsInitialized()).To(BeFalse())
+			err = hc.RemoveTaint(TAINT_KEY)
+			Expect(err).To(HaveOccurred())
+			Expect(hc.IsInitialized()).To(BeFalse())
+		})
+		var _ = It("return error when trying to remove taint (update node)", func() {
+			c := fake.NewSimpleClientset(fakeNodes)
+			c.CoreV1().(*fakecorev1.FakeCoreV1).PrependReactor("update", "nodes",
```
schrej, line 128: controller-runtime also has a pretty nice fakeclient if you are able to use that one

pkg/reconciler/reconciler.go
```diff
@@ -57,6 +62,29 @@ func NewReconciler(client client.Client, anycastTracker *anycast.AnycastTracker)
 		reconciler.config = config
 	}
 
+	cfg, err := clientgorest.InClusterConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	c, err := clientgo.NewForConfig(cfg)
```
schrej, line 70: same as above, consider using the controller-runtime client

schrej (changes requested):

main.go
```diff
@@ -159,6 +162,47 @@ func main() {
...
 	}
 
+	if onlyBPFMode {
+		cfg, err := clientgorest.InClusterConfig()
+		if err != nil {
+			setupLog.Error(err, "problem creating cluster config")
+			os.Exit(1)
+		}
+
+		c, err := clientgo.NewForConfig(cfg)
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Yes, 'veI tried to use it, but it seems that controller-runtime's client cannot be used before mgr.Start() call.
When I try to use it in main function (for bpf-only node) it returns
`{"error": "the cache is not started, can not read objects"}` when trying to get the node's object from the API.

pkg/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,211 @@
...
+	toolkit   *HealthCheckToolkit
+}
+
+func NewHealthChecker(client clientgo.Interface, toolkit *HealthCheckToolkit, netconf *NetHealthcheckConfig) (*HealthChecker, error) {
+	ctx := context.Background()
+	return &HealthChecker{
+		client:        client,
+		ctx:           ctx,
+		isInitialized: false,
+		Logger:        log.FromContext(ctx),
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Changed that to `log.Log.WithName("HealthCheck")`

pkg/healthcheck/healthcheck_test.go
p-strusiewiczsurmacki-mobica: OK, I've removed my mocks and uber-go/mock is now used.

pkg/healthcheck/healthcheck_test.go
p-strusiewiczsurmacki-mobica: Done

pkg/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,211 @@
...
+		configFile = val
+	}
+
+	read, err := os.ReadFile(configFile)
+	if isMandatory && err != nil {
+		return nil, err
+	}
+	if !isMandatory && err != nil {
+		return config, nil
+	}
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Reworked that. Added additional bool variable and made the if easier to understand, I think.

pkg/healthcheck/healthcheck_test.go
schrej: Would be good to add a `// go:generate <gomock command>` somewhere, maybe in proximity of those interfaces. Then we can just run `go generate ./...` to regenerate if necessary

main.go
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Done

pkg/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,211 @@
...
+	if link.Attrs().OperState != netlink.OperUp {
+		return errors.New("link " + intf + " is not up - current state: " + link.Attrs().OperState.String())
+	}
+	return nil
+}
+
+func (hc *HealthChecker) CheckReachability() error {
+	for _, i := range hc.netConfig.Reachability {
+		if err := hc.checkReachabilityItem(i); err != nil {
+			if strings.Contains(err.Error(), "refused") {
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Done

pkg/healthcheck/healthcheck_test.go
```diff
@@ -0,0 +1,355 @@
...
+	RegisterFailHandler(Fail)
+	t.Setenv(NODE_NAME_ENV, TESTHOSTNAME)
+	tmpPath = t.TempDir()
+	RunSpecsWithDefaultAndCustomReporters(t,
+		"HealthCheck Suite",
+		[]Reporter{printer.NewlineReporter{}})
+}
+
+var _ = Describe("HealthCheck", func() {
+	var _ = Context("LoadConfig() should", func() {
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: I believe that I've seen linter that was not happy about unhandled return values before. Anyway, changed that as I can't see similar issue with current golangci-lint setup.

pkg/healthcheck/healthcheck_test.go
p-strusiewiczsurmacki-mobica: Done

Did not want to add irrelevant files to the repository, but if you say it's OK, then it's simpler approach, which I like.

pkg/healthcheck/healthcheck_test.go
p-strusiewiczsurmacki-mobica: Done

pkg/healthcheck/healthcheck_test.go
p-strusiewiczsurmacki-mobica: Done

Although, had to add a mock for this client's error, as can't find anything similar to `PrependReactor` available in goclient.

pkg/reconciler/reconciler.go
p-strusiewiczsurmacki-mobica: Done.

Used reconciler.client for this one as it's already created.

main.go
```diff
@@ -159,6 +162,41 @@ func main() {
 		macvlan.RunMACSync(interfacePrefix)
 	}
 
+	if onlyBPFMode {
+		clusterClient, err := client.New(clientConfig, client.Options{})
+		if err != nil {
+			setupLog.Error(err, "problem kreating controller-runtime client")
```
schrej, line 168:
```suggestion
			setupLog.Error(err, "problem creating controller-runtime client")
```

pkg/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+	return &net.Dialer{Timeout: timeout}
+}
+
+// NewDefaultHealthcheckToolkit returns.
+func NewDefaultHealthcheckToolkit(frr frrInterface, tcpDialer tcpDialerInterface) *Toolkit {
+	return NewHealthCheckToolkit(frr, netlink.LinkByName, tcpDialer)
+}
+
+//go:generate mockgen -source=../../pkg/healthcheck/healthcheck.go -destination=../../pkg/healthcheck/mock/mock_healthcheck.go
```
schrej, line 233: Can you try the following here?
```
//go:generate mockgen -destination ../mock/mock_healtheck.go . frrInterface,tcpDialerInterface
```
alternatively this could be in something like a `doc.go` file in the `mock` package so the command is closer to the generated code. In that case the `.` would need to be replaced by the full package path (`github.com/telekom/das-schiff-network-operator/pkg/healthcheck`)

schrej (approved):
Two more nits, the rest looks good from my side, thank you!

I'll leave it to Christopher for a second review.

pkg/healthcheck/healthcheck.go
p-strusiewiczsurmacki-mobica: Done. :) Had to export the interfaces though.

chdxD1 (commented):
LGTM so far. I would like to have some kind of retry, especially for the connectivity step.

Also what I forgot: We need to come up with a way to specify interfaces names per node because they might differ (usually on masters we have eno1, eno2, on workers the PCIe cards might be on different slots, depending on HW type, resulting in different ens<slot> names)

p-strusiewiczsurmacki-mobica:
@chdxD1 
> LGTM so far. I would like to have some kind of retry, especially for the connectivity step.

Just added retries for the reachability test. Updated readme so you can quickly check if it's what you were thinking about. Additionally fixed a bug I found in the reachability test.

Should I add some sleep between retries?

> Also what I forgot: We need to come up with a way to specify interfaces names per node because they might differ (usually on masters we have eno1, eno2, on workers the PCIe cards might be on different slots, depending on HW type, resulting in different ens names)

As for the interfaces names - it is currently configured per node as the config file is separate for each node. But it's not the most user-friendly way I think. Maybe some configmaps would be better? If so, maybe it could be done in another PR?

chdxD1:
@p-strusiewiczsurmacki-mobica missed that config per node thing, I initially assumed that it was part of the overall config. It is probably fine for now, other solutions would be a bit more complicated. We can provide the config as part of our deployment process!

README.md
```diff
@@ -87,6 +87,39 @@ An internal loop is tracking these interfaces and reapplies the eBPF code when n
...
 ![eBPF Flow](docs/ebpf-flow.png)
 
+### Networking healthcheck
+
+After deployment basic networking connectivity can be tested to ensure that the node is fully operable. If all checks will pass, taint `node.cloudprovider.kubernetes.io/uninitialized` will be removed from the node if applied.
+
+There are 3 types of checks implemented currently:
+ - FRR check - will check if FRR's daemon is up and running. This check will be only used in pods that DO NOT work in bpf-only mode.
+ - interfaces - will check if all network interfaces that are expected to are up
+ - reachability - will check if TCP connection to provided hosts can be achieved. Each host will be queired `retries` times before reporting failure.
```
watzkuh, line 97:
```suggestion
 - reachability - will check if TCP connection to provided hosts can be achieved. Each host will be queried `retries` times before reporting failure.
```

README.md
```diff
@@ -87,6 +87,39 @@ An internal loop is tracking these interfaces and reapplies the eBPF code when n
...
+
+ Interfaces and reachability checks can be configured using config file located in `/opt/network-operator/net-healthcheck-config.yaml` on the node.
```
watzkuh, line 99:
```suggestion
 Interfaces and reachability checks can be configured through a config file located in `/opt/network-operator/net-healthcheck-config.yaml` on the node.
```

README.md
```diff
@@ -87,6 +87,39 @@ An internal loop is tracking these interfaces and reapplies the eBPF code when n
...
+ ```
+interfaces: // list of interfaces that should have UP state
+  - eth0
+reachability: // list of hosts and ports to query in reachability check
+- host: "host.domain.com"
+  port: 80
+- host: "192.168.1.100"
+  port: 8400
+timeout: 5s // timeout for reachability test (if not provided, default 3s will be used)
+retries: 10 // idicates how many times each host provided in reachibility list will be queried before failure will be reported, default 3
```
watzkuh, line 112:
```suggestion
interfaces: # list of interfaces that should have UP state
  - eth0
reachability: # list of hosts and ports to query in reachability check
- host: "host.domain.com"
  port: 80
- host: "192.168.1.100"
  port: 8400
timeout: 5s # timeout for reachability test (if not provided, default 3s will be used)
retries: 10 # indicates how many times each host provided in reachibility list will be queried before failure will be reported, default 3
```

README.md
```diff
@@ -87,6 +87,39 @@ An internal loop is tracking these interfaces and reapplies the eBPF code when n
...
+ ```
+
+Location of the networking healthcheck config file can be altered by setting `OPERATOR_NETHEALTHCHECK_CONFIG` enviromental vairable on the node to selected path, e.q.
```
watzkuh, line 115:
```suggestion
The location of the networking healthcheck config file can be altered by setting the `OPERATOR_NETHEALTHCHECK_CONFIG` environment variable on the node to the desired path, e.g.:
```

README.md
```diff
@@ -87,6 +87,39 @@ An internal loop is tracking these interfaces and reapplies the eBPF code when n
...
+
+```bash
+OPERATOR_NETHEALTHCHECK_CONFIG=/etc/network-operator/my-healtcheck-config.yaml
+```
+
+> NOTE: Please be aware that custom config file location has to be mounted into the network-operator pod.
```
watzkuh, line 121:
```suggestion
> NOTE: Please be aware that the custom config file location has to be mounted into the network-operator pod.
```

pkg/healthcheck/healthcheck.go
```diff
@@ -0,0 +1,269 @@
...
+
+// NewTCPDialer returns new tcpDialerInterface.
+func NewTCPDialer(dialerTimeout string) TCPDialerInterface {
+	timeout, err := time.ParseDuration(dialerTimeout)
+	if err != nil {
+		seconds, err := strconv.Atoi(dialerTimeout)
+		if err != nil {
+			timeout = time.Second * defaultTCPTimeout
+		} else {
+			timeout = time.Second * time.Duration(seconds)
```
watzkuh, line 251: How hard to we want to try to "silently do the right thing" here?
Might be worth to at least log a warning that we can't deal with the config the user provided (and will now just assume they probably meant seconds / fallback to the default timeout).

pkg/reconciler/reconciler.go
```diff
@@ -86,5 +101,22 @@ func (reconciler *Reconciler) reconcileDebounced(ctx context.Context) error {
 	if err := r.reconcileLayer2(l2vnis); err != nil {
 		return err
 	}
+
+	if !reconciler.healthChecker.IsInitialized() {
```
watzkuh, line 105: Why are we doing all these checks if the healthChecker is not yet initialized?

Wouldn't we expect them to fail anyway?

watzkuh:
LGTM besides those very minor comments

pkg/reconciler/reconciler.go
p-strusiewiczsurmacki-mobica: I think the method name was not descriptive enough. Changed it to `IsNetworkingHealthy()`.

The whole point of this check is, because reconciler runs in a loop, to run the healthcheck only if networking was not already checked and determined to be healthy, and set `isNetworkingHealhy` (formerly `isInitialized`) variable accordingly. After it is set, node taint should be removed and healthcheck should not be performed again.

pkg/healthcheck/healthcheck.go
p-strusiewiczsurmacki-mobica: > How hard to we want to try to "silently do the right thing" here? 

I'm completely open to suggestions on that. Maybe others would like to give their insights?

Anyway, added logging for now to give user more info on what's going on. That's really good suggestion, btw!

Cellebyte (commented):
LGTM 

added hint that it closes the Issue raised regarding node discovery by hostname.

PR #48, opened 2023-09-07
Fixed logging of healthcheck.go file

Cellebyte:
* wrong call to logging function `Info` which caused completely broken network-operator.
* fixed in #4 and cherry-picked here.


watzkuh:
LGTM

PR #63, opened 2023-09-27
Added separate go entrypoints for frr-monitoring and the manager

Cellebyte:
* Separate frr-monitoring from the rest of the operator.
* This was needed as the normal network-operator image does not provide a vtysh for the interaction with frr.
* The main.go was moved into the `./cmd` directory which is an unwrittten standard for multi-cli projects -> [https://github.com/golang-standards/project-layout/tree/master/cmd](https://github.com/golang-standards/project-layout/tree/master/cmd)
* Also adds timeouts for the prometheus handler -> [https://app.deepsource.com/directory/analyzers/go/issues/GO-S2114](https://app.deepsource.com/directory/analyzers/go/issues/GO-S2114)


cmd/frr-monitoring/main.go
```diff
@@ -0,0 +1,58 @@
+package main
```
schrej, line 1: Do we want to call this `frr-exporter` to be more in line with common naming in the prometheus ecosystem? Or is that not accurate?

Cellebyte: renamed it to frr-exporter.

chdxD1 (approved):

PR #65, opened 2023-10-10
Wait for FRR reload to complete, restart if failed

chdxD1:
We now wait for the FRR reload to complete (waiting for a message on the channel), if the reloadStatus (aka job status) is not `done` we trigger an FRR restart.


Cellebyte (commented):
LGTM
First Reload and if not working restart should be enough.

Logging that restart happened could also be beneficial.

Cellebyte (approved):
LGTM

PR #67, opened 2023-10-16
Template aggregates into MGMT VRF (aka skip template VRFs)

chdxD1:
In case of the management VRF we currently do not support specifying aggregates. This MR will allow addition of Aggregate addresses from VRFs that are configured to be skipped.


Cellebyte (approved):
LGTM

PR #68, opened 2023-10-17
Actually pass Community into RouteMap for configuration

chdxD1:


Cellebyte (approved):
Nice finally the communities are also used ^^

PR #70, opened 2023-10-17
Re-Attach/Update BPF program on existing VRF interfaces and Fix Monitoring

chdxD1:
Previously the BPF program was only re-attached to the onlyBPFInterfaces of the config. Existing VRF interfaces weren't reconciled and the BPF wasn't re-added to them. This is now fixed and the BPF program is re-attached if the IDs do not match.

In this PR also the monitoring of eBPF maps get fixed by copying the value of the for-loop into a separate variable, which is a known Go problem (which gets fixed in 1.21 which we don't use yet)

### DIFF
--- a/pkg/monitoring/bpf.go
+++ b/pkg/monitoring/bpf.go
@@ -2,6 +2,7 @@ package monitoring
 
 import (
 	"fmt"
+
 	"github.com/cilium/ebpf"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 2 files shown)


Cellebyte:
I would suggest moving the bpf monitoring into the monitoring directory.

schrej (approved):
lgtm code wise

chdxD1:
Confirmed working on dev Cluster 🙌

PR #86, opened 2023-12-20
Add Routing Table as a service CRD

chdxD1:


controllers/routingtable_controller.go
```diff
@@ -0,0 +1,70 @@
...
+//+kubebuilder:rbac:groups=network.schiff.telekom.de,resources=routingtables,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=network.schiff.telekom.de,resources=routingtables/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=network.schiff.telekom.de,resources=routingtables/finalizers,verbs=update
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the RoutingTable object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
```
watzkuh, line 48:
```suggestion
```
Or is there something left to do here?

pkg/nl/taas.go
```diff
@@ -0,0 +1,31 @@
...
+
+func (n *NetlinkManager) CreateTaas(info TaasInformation) error {
+	_, err := n.createVRF(info.Name, info.Table)
+	if err != nil {
+		return fmt.Errorf("error creating VRF for TaaS: %w", err)
+	}
+
+	err = n.setUp(info.Name)
+	if err != nil {
+		return fmt.Errorf("error set VRF up for TaaS: %w", err)
```
watzkuh, line 18:
```suggestion
		return fmt.Errorf("error setting up VRF for TaaS: %w", err)
```

pkg/reconciler/layer3.go
```diff
@@ -283,7 +283,7 @@ func (r *reconcile) cleanupTaasNetlink(existing []nl.TaasInformation, intended [
 	for _, cfg := range existing {
 		stillExists := false
 		for i := range intended {
-			if intended[i].IsTaaS && intended[i].Name == cfg.Name && intended[i].VNI == cfg.Table {
+			if intended[i].Name == cfg.Name && intended[i].VNI == cfg.Table {
```
Cellebyte, line 306: and this `cfg.Table == intended[i].VNI works` is there not a limit on table ids?

pkg/reconciler/layer3.go
```diff
@@ -63,21 +63,22 @@ func (r *reconcile) reconcileLayer3(l3vnis []networkv1alpha1.VRFRouteConfigurati
...
 		r.Logger.Error(err, "error reconciling Netlink")
 		return err
 	}
 
-	err = r.reconcileTaasNetlink(taasConfigs)
+	deletedTaas, err := r.reconcileTaasNetlink(taasConfigs)
 	if err != nil {
 		return err
 	}
+	reloadTwice := deletedVRF || deletedTaas
```
Cellebyte, line 76: why do we need to reload twice when a vrf is removed?

pkg/reconciler/layer3.go
chdxD1: I am using VNI as the field for tableID as well, limit is 2^32-1, it's a 32bit uint

pkg/reconciler/layer3.go
chdxD1: First reload removes all statements but does not remove `router bgp <asn> vrf <vrf>` statement. This is achieved by a second reload.

pkg/reconciler/layer3.go
Cellebyte: ok. then it fits :)

pkg/reconciler/layer3.go
Cellebyte: 👍

pkg/reconciler/layer3.go
```diff
@@ -123,6 +175,24 @@ func (r *reconcile) createVrfConfigMap(l3vnis []networkv1alpha1.VRFRouteConfigur
 	return vrfConfigMap, nil
 }
 
+func createVrfFromTaaS(taas []networkv1alpha1.RoutingTable) map[string]frr.VRFConfiguration {
+	vrfConfigMap := map[string]frr.VRFConfiguration{}
+
+	for i := range taas {
+		spec := taas[i].Spec
+
+		name := fmt.Sprintf("taas.%d", spec.TableID)
```
Cellebyte, line 184: this could break hardly in BM4X if we don't extend the interface naming in linux.

Cellebyte: Ahh I just miscalculated it ^^

Cellebyte (approved):
LGTM

PR #88, opened 2024-01-04
Check if duplicate IDs are used for Layer2NetworkConfiguration

chdxD1:


Cellebyte:
should we not block that on apply?

pkg/reconciler/layer2.go
```diff
@@ -69,6 +69,10 @@ func (r *reconcile) fetchLayer2(ctx context.Context) ([]networkv1alpha1.Layer2Ne
 		l2vnis = append(l2vnis, *item)
 	}
 
+	if err := r.checkL2Duplicates(l2vnis); err != nil {
```
Cellebyte, line 72: should we not block mismatching resources before they are in cluster?

Cellebyte: Ok :)

watzkuh:
lgtm

PR #89, opened 2024-01-04
Implement VRF MTU

chdxD1:


pkg/nl/layer3.go
```diff
@@ -164,3 +165,33 @@ func (*NetlinkManager) EnsureBPFProgram(info VRFInformation) error {
 
 	return nil
 }
+
+func (info VRFInformation) linkMTU() int {
+	if info.MTU == 0 {
+		return defaultMtu
```
schrej, line 171: Could also be done using a defaulting webhook so it's also represented in the CRD. Not mandatory though, it might be excessive to have one just for that.

pkg/nl/layer3.go
```diff
@@ -164,3 +165,33 @@ func (*NetlinkManager) EnsureBPFProgram(info VRFInformation) error {
...
+	}
+	if link.Attrs().MTU != info.linkMTU() {
+		if err := netlink.LinkSetMTU(link, info.MTU); err != nil {
+			return fmt.Errorf("error setting MTU of vrf2default interface of vrf %s: %w", info.Name, err)
+		}
+	}
+
+	link, err = netlink.LinkByName(defaultToVrfPrefix + info.Name)
+	if err != nil {
+		return fmt.Errorf("error getting vrf2default interface of vrf %s: %w", info.Name, err)
```
schrej, line 189: should probably be `default2vrf` here

schrej (approved):
one comment, one nit

pkg/nl/layer3.go
chdxD1: it's already set as a default in the CRD and that apparently sets it in the cluster as well (without using a webhook)

Issue #96, opened 2024-02-15
Extended Visibility for Host Routing


chdxD1:
As of today we have to connect to all nodes one-by-one, open `vtysh` and run debug commands. This is a rather tedious process.

I'd like to propose an API endpoint on each of the monitoring containers (where it is probably best suited because we already have a vtysh connection and so on) that gathers this information on each node.

In addition to that there should be another **endpoint** that allows retrieving information from **all nodes** in a cluster by querying just one network-operator pod.

List of commands that should be implemented (all output can be returned as is, without any sanitization):
- `show ip/ipv6 route (vrf <vrf>) <input> json` (with/without longer-prefixes)
- `show bgp (vrf <vrf>) ipv4/ipv6 unicast <input> json` (with/without longer-prefixes)
- `show evpn vni json`
- `show evpn rmac vni <all|vrf> json`
- `show evpn mac vni <all|vrf> json`
- `show evpn next-hops vni <all|vrf> json`
- `show bgp vrf <all|vrf> summary json`

p-strusiewiczsurmacki-mobica:
I was finally able to start working on this. Just created WIP PR.

For now there are 3 GET endpoints.

**Endpoint /show/route:**

- show ip/ipv6 route (vrf <vrf>) <input> json (with/without longer-prefixes)
- 
Example: /show/route?protocol=<ip/ipv6>&vrf=<vrf>

Can you tell me what <input> is and how to switch between `with/without longer-prefixes`. Or point me to the docs, as I can't find this.

**Endpoint: /show/bgp**

- show bgp (vrf <vrf>) ipv4/ipv6 unicast <input> json (with/without longer-prefixes)

Example: /show/bgp?vrf=<vrf>&protocol=<ipv4/ipv6>

Same question about <input> and prefixes applies here. Also, will

show bgp vrf **all** ipv4/ipv6 unicast <input> json

be a valid command?


- show bgp vrf <all|vrf> summary json

Example: /show/bgp?vrf=<vrf>&type=summary

**Endpoint /show/evpn:**
- show evpn vni json

Example: /show/evpn

- show evpn rmac vni <all|vrf> json

Example: /show/evpn?type=rmac&vrf=<vrf>

- show evpn mac vni <all|vrf> json
 
Example: Example: /show/evpn?type=mac&vrf=<vrf>

- show evpn next-hops vni <all|vrf> json

Example: Example: /show/evpn?type=next-hops&vrf=<vrf>

[...]

PR #100, opened 2024-02-28
Added monitoring endpoint

p-strusiewiczsurmacki-mobica:
This PR adds monitoring endpoints as described in #96.

- show ip/ipv6 route (vrf \<vrf\>) \<input\> json (with/without longer-prefixes)
- show bgp (vrf \<vrf\>) ipv4/ipv6 unicast \<input\> json (with/without longer-prefixes)
- show evpn vni json
- show evpn rmac vni \<all|vrf\> json
- show evpn mac vni \<all|vrf\> json
- show evpn next-hops vni \<all|vrf\> json
- show bgp vrf \<all|vrf\> summary json


pkg/monitoring/endpoint.go
```diff
@@ -0,0 +1,113 @@
...
+func (e *Endpoint) ShowRoute(w http.ResponseWriter, r *http.Request) {
+	vrf := r.URL.Query().Get("vrf")
+	if vrf == "" {
+		vrf = vrfAll
+	}
+
+	protocol := r.URL.Query().Get("protocol")
+	if protocol == "" {
+		protocol = "ip"
+	}
```
chdxD1, line 39: Please check if the protocol makes sense: ip and ipv6

pkg/monitoring/endpoint.go
```diff
@@ -0,0 +1,113 @@
...
+			"bgp",
+			"vrf",
+			vrf,
+			"summary",
+		})
+	} else {
+		protocol := r.URL.Query().Get("protocol")
+		if protocol == "" {
+			protocol = "ipv4"
+		}
```
chdxD1, line 73: protocol is either ipv4 unicast or ipv6 unicast. You can map the protocol query parameter like this:
- ip --> ipv4 unicast
- ipv6 --> ipv6 unicast

pkg/monitoring/endpoint.go
```diff
@@ -0,0 +1,113 @@
...
+			"unicast",
+		})
+	}
+
+	writeResponse(&data, w)
+}
+
+func (e *Endpoint) ShowEVPN(w http.ResponseWriter, r *http.Request) {
+	data := []byte{}
+	requestType := r.URL.Query().Get("type")
```
chdxD1, line 90: requestType should also be checked against the list of allowed types: rmac, mac, next-hops

pkg/monitoring/endpoint.go
```diff
@@ -0,0 +1,113 @@
...
+		data = e.cli.ExecuteWithJSON([]string{
+			"show",
+			"evpn",
+			"vni",
+		})
+	} else {
+		vrf := r.URL.Query().Get("vrf")
+		if vrf == "" {
+			vrf = vrfAll
+		}
```
chdxD1, line 101: vrf is actually `vni` in this special case (it is a number but `all` is supported as well). I mistyped that in my original issue

chdxD1 (changes requested):

pkg/monitoring/endpoint.go
p-strusiewiczsurmacki-mobica: Done

pkg/monitoring/endpoint.go
p-strusiewiczsurmacki-mobica: Done.

pkg/monitoring/endpoint.go
p-strusiewiczsurmacki-mobica: Changed that a bit. Can you check if it is OK now?

pkg/monitoring/endpoint.go
```diff
@@ -0,0 +1,307 @@
...
+	matchLabels := &client.MatchingLabels{
+		"app.kubernetes.io/name":      "network-operator",
+		"app.kubernetes.io/component": "worker",
+	}
+
+	err := e.c.List(r.Context(), pods, matchLabels)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("error listing pods: %s", err.Error()), http.StatusInternalServerError)
+		return
+	}
```
chdxD1, line 201: I am not that happy about the pod discovery, I would probably go for a service / service endpoint based discovery. That way we can pass a servicename as a parameter to the pod.

pkg/monitoring/endpoint.go
```diff
@@ -0,0 +1,307 @@
...
+
+	for err := range errors {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	for result := range results {
+		buffer = append(buffer, result...)
+		buffer = append(buffer, '\n')
+	}
```
chdxD1, line 239: All the results should be in JSON format, you can probably parse them into `map[string]interface{}` individually. Afterwards you can use a map of that again (hostname to answer parsed into map[string]interface{}).

This way you get meaningful JSON from the aggregate endpoint without relying on string concatenation.

chdxD1 (changes requested):

pkg/monitoring/endpoint.go
p-strusiewiczsurmacki-mobica: You're right. Changed this to use `map[string]json.RawMessage{}` for proper marshaling.

pkg/monitoring/endpoint.go
p-strusiewiczsurmacki-mobica: I've changed that. Now the `/all` endpoints will accept parameters `service` and `namespace`, eg.
`/all/show/routes?service=some-svc&namespace=some-namespace`
It will then try to find service and `endpointslices` for this service. If endpointslices are not available it will fallback to the legacy `endpoints`.
I am not sure how this should work with dualstack clusters though. For now I just use the first endpointslices/endpoints object that matches the service.

Cellebyte: If you want some example of how to the discovery you can read a little bit here :) https://github.com/Cellebyte/kube-latency/blob/master/app.go#L192

Cellebyte: This is using dns discovery, but as @chdxD1 already stated servicename and namespace could be passed on deployment so it is not needed to have them in the URL :)

p-strusiewiczsurmacki-mobica: @Cellebyte Thank you, that's much cleaner. :)
@chdxD1 Service is now configured using `STATUS_SVC_NAME` and `STATUS_SVC_NAMESPACE` variables. Added those to the manager deployment definitions, and added NodePort service `network-operator-status` definition to the deployment.

config/manager/service.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: status
+  namespace: system
+spec:
+  type: NodePort
```
chdxD1, line 7: Service should be ClusterIP and .spec.clusterIP should be set to none: https://kubernetes.io/docs/concepts/services-networking/service/#headless-services

config/manager/service.yaml
```diff
@@ -0,0 +1,15 @@
...
+  selector:
+    app.kubernetes.io/component: worker
+    app.kubernetes.io/name: network-operator
+  ports:
+    - protocol: TCP
+      port: 7082
+      targetPort: 7082
+      nodePort: 30082
```
chdxD1, line 15: remove nodePort from here

config/rbac/role.yaml
```diff
@@ -13,6 +13,18 @@ rules:
   - list
   - update
   - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - list
```
chdxD1, line 21: we shouldn't need pods anymore

config/rbac/role.yaml
```diff
@@ -13,6 +13,18 @@ rules:
...
+- apiGroups:
+  - ""
+  resources:
+  - services
+  verbs:
```
chdxD1, line 26: we might need endpoints & endpoint slices in addition to pods though if we go with service based discovery

pkg/monitoring/endpoint.go
```diff
@@ -0,0 +1,388 @@
...
+	}); err != nil {
+		return nil, fmt.Errorf("error getting pods: %w", err)
+	}
+
+	addresses := []string{}
+	for i := range pods.Items {
+		addresses = append(addresses, pods.Items[i].Status.PodIP)
+	}
+
+	return addresses, nil
```
chdxD1, line 337: oh I see, this is not using the actual service endpoint(slice)s but using the pods from the labelSelector?

config/manager/service.yaml
p-strusiewiczsurmacki-mobica: Done

config/manager/service.yaml
p-strusiewiczsurmacki-mobica: Done

config/rbac/role.yaml
p-strusiewiczsurmacki-mobica: As you've seen in your next comment [we still list pods here](https://github.com/p-strusiewiczsurmacki-mobica/das-schiff-network-operator/blob/extended-routing-visability/pkg/monitoring/endpoint.go#L326)

config/rbac/role.yaml
p-strusiewiczsurmacki-mobica: Not using enpoints and enpointslices currently, as you've stated in next comment here.

pkg/monitoring/endpoint.go
p-strusiewiczsurmacki-mobica: Yes, I had it implemented using enpoints and endpointslices but @Cellebyte suggested [another approach in this comment](https://github.com/telekom/das-schiff-network-operator/pull/100#discussion_r1525377202), which I think is much more clean, so I went this way. If you think it would be better to use enpoints and enpointslices directly for any reason I can revert that.

pkg/monitoring/endpoint.go
```diff
@@ -0,0 +1,389 @@
...
+}
+
+// SetHandlers configures HTTP handlers.
+func (e *Endpoint) SetHandlers() {
+	http.HandleFunc("/show/route", e.ShowRoute)
+	http.HandleFunc("/show/bgp", e.ShowBGP)
+	http.HandleFunc("/show/evpn", e.ShowEVPN)
+	http.HandleFunc("/all/show/route", e.PassRequest)
+	http.HandleFunc("/all/show/bgp", e.PassRequest)
+	http.HandleFunc("/all/show/evpn", e.PassRequest)
```
schrej, line 56: Can we please create our own http server instance? I saw that it's also using the default instance in main.go for the frr-exporter, please use a custom instance for everything.

pkg/monitoring/endpoint.go
```diff
@@ -0,0 +1,389 @@
...
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	if err := setLongerPrefixes(r, &command); err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	data := e.cli.ExecuteWithJSON(command)
```
schrej, line 93: Is this safe? We're directly passing query args (vrf) to the command. Do we need to do some additional sanitization here to prevent RCE? I think making sure it's only allowed characters would go a long way.

Same applies to all other cases.

pkg/monitoring/endpoint.go
```diff
@@ -0,0 +1,389 @@
...
+		return
+	}
+
+	writeResponse(result, w)
+}
+
+//+kubebuilder:rbac:groups=core,resources=services,verbs=get
+
+// PassRequest - when called, will pass the request to all nodes and return their responses.
+func (e *Endpoint) PassRequest(w http.ResponseWriter, r *http.Request) {
```
schrej, line 212: Maybe `QueryAll` instead

pkg/monitoring/endpoint.go
```diff
@@ -0,0 +1,389 @@
...
+	query := strings.ReplaceAll(r.URL.RequestURI(), "all/", "")
+	responses := []json.RawMessage{}
+
+	var wg sync.WaitGroup
+	results := make(chan []byte, len(addr))
+	errors := make(chan error, len(addr))
+
+	for i := range addr {
+		wg.Add(1)
+		go passRequest(r, addr[i], query, results, errors, &wg)
```
schrej, line 351: You could also do this so it's a little bit more separated
```suggestion
		go func() {
		  passRequest(r, addr[i], query, results, errors, &wg)
		  wg.Done()
		}()
```

pkg/monitoring/endpoint.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+
+	wg.Wait()
+	close(results)
+	close(errors)
+
+	for err := range errors {
+		return nil, fmt.Errorf("error occurred: %w", err)
+	}
```
schrej, line 360: This will only return the first error, so `if len(errors) > 0 { return nil, errors[0] }` would be more efficient. Always a bit tricky to do this properly, maybe just return a list of errors for now.

In general I've been thinking it might make sense to add proper logging, and when doing that hide a few more errors in the logs instead of returning them in the response.

schrej (changes requested):

pkg/monitoring/endpoint.go
p-strusiewiczsurmacki-mobica: I've created separate server custom instances for both monitoring endpoint and frr-exporter.
However, adding another instance would require one of those to be run in separate goroutine which would require some sort of communication between those two (for example, if one server dies, the other one should be stopped gracefully and the pod/container should be restarted, etc). I think this is not really worth it as it makes this unnecessary complex, so I've created new `cmd/monitoring` directory and added new `monitoring` image that is run in separate container in the network-operator pod.
Do you think this will be OK, or should we keep frr-exporter and monitoring in the same image/container?

pkg/monitoring/endpoint.go
p-strusiewiczsurmacki-mobica: Done

pkg/monitoring/endpoint.go
p-strusiewiczsurmacki-mobica: Changed that. If multiple errors will occur, the response will be
```
http.Error(w, "multiple errors occurred while querying endpoints - please check logs for the details", http.StatusInternalServerError)
```

Also added some logging, which looks like this for now:

```
2024-04-18T14:51:06Z	INFO	setup	loaded kubernetes config
2024-04-18T14:51:06Z	INFO	setup	loaded status service config
2024-04-18T14:51:06Z	INFO	monitoring	created ServeMux
2024-04-18T14:51:06Z	INFO	setup	created server, starting...	{"Addr": ":7083", "ReadHeaderTimeout": "20s", "ReadTimeout": "1m0s"}
2024-04-18T14:51:24Z	INFO	monitoring	got ShowRoute request
2024-04-18T14:51:24Z	INFO	monitoring	command to be executed	{"command": ["show", "ip", "route", "vrf", "all"]}
2024-04-18T14:51:25Z	INFO	monitoring	response written	{"type": "ShowRoute"}
2024-04-18T14:51:42Z	INFO	monitoring	got QueryAll request
2024-04-18T14:51:42Z	INFO	monitoring	will querry endpoints	{"endpoints": ["172.18.0.3", "172.18.0.6", "172.18.0.8", "172.18.0.7", "172.18.0.5", "172.18.0.2"]}
2024-04-18T14:51:42Z	INFO	monitoring	got ShowRoute request
2024-04-18T14:51:42Z	INFO	monitoring	command to be executed	{"command": ["show", "ip", "route", "vrf", "all"]}
2024-04-18T14:51:42Z	INFO	monitoring	response written	{"type": "ShowRoute"}
2024-04-18T14:51:42Z	INFO	monitoring	response written	{"type": "QueryAll"}
```

pkg/monitoring/endpoint.go
p-strusiewiczsurmacki-mobica: Added that. However had to change it slightly as it seems that `i` variable is a reference, so decided to pass it by value to the lambda function just to be sure it won't get changed outside the function.

pkg/monitoring/endpoint.go
schrej: Ugh, I'm sorry, I should've been clearer here. I didn't mean to split it into two http servers, I just wanted to not use the default http server. I even thought about it might being a bit to unclear but then decided it's good enough...
Can you change it back to a single server instance? Just don't use the default mux of the `http` package.

p-strusiewiczsurmacki-mobica: No problem. :) In fact, that's my bad. I've should have thought that by 'a custom instance' (not instance**s** 😅) you've meant just one.
Anyway, it should be fixed now. :)

schrej (approved):
lgtm, please wait for feedback on that validation comment from @chdxD1

pkg/monitoring/endpoint.go
chdxD1: dash and underscore should both be valid parameters here. VNI is a 24-bit number.

p-strusiewiczsurmacki-mobica: OK, changed VRF regexp to
"^[a-zA-Z0-9-_.]+$"

And added validation function for VNI: https://github.com/telekom/das-schiff-network-operator/blob/0ba752b0ad498317a52064e522aaa1c125bbf2db/pkg/monitoring/endpoint.go#L266

PR #105, opened 2024-03-15
Prohibit usage of vni zero

jetzorkj:


pkg/reconciler/layer3.go
```diff
@@ -165,6 +165,12 @@ func (r *reconcile) createVrfConfigMap(l3vnis []networkv1alpha1.VRFRouteConfigur
 			continue
 		}
 
+		if vni == 0 {
```
Cellebyte, line 168: Should we not do a boundary check here? 

`vni <= 0 and vni >= max_vni`

MaxRink:
I mean the apiserver can also do that enforcement with a block directly

https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/

Marcel Fest ***@***.***> schrieb am Fr., 15. März 2024, 21:12:

Cellebyte:
@MaxRink only for the CRD ^^ the ConfigMap needs validation as well :)

MaxRink:
Make the ConfigMap an CRD :P


Am Sa., 16. März 2024 um 11:25 Uhr schrieb Marcel Fest <
***@***.***>:

config/crd/bases/network.schiff.telekom.de_layer2networkconfigurations.yaml
```diff
@@ -126,6 +126,7 @@ spec:
                 x-kubernetes-map-type: atomic
               vni:
                 description: VXLAN VNI Id for the layer 2 network
+                minimum: 1
```
Cellebyte, line 129: we should add the `maximum: 16777215` as well to also validate the upper boundary.

pkg/reconciler/layer3.go
```diff
@@ -165,6 +165,12 @@ func (r *reconcile) createVrfConfigMap(l3vnis []networkv1alpha1.VRFRouteConfigur
 			continue
 		}
 
+		if vni == 0 && vni > 16777215 {
+			err := fmt.Errorf("VNI can not be set to 0")
```
Cellebyte, line 169: please also add the upper boundary in the error message

api/v1alpha1/layer2networkconfiguration_types.go
```diff
@@ -37,6 +37,7 @@ type Layer2NetworkConfigurationSpec struct {
 	MTU int `json:"mtu"`
 
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=1
```
Cellebyte, line 40: you need to regenerate the types again

Issue #98, opened 2024-02-20
Gradual Rollout Support


chdxD1:
In the current design network-operator runs independently from each other. Custom resources are read from the cluster and configured on the local node.

If a user configures a faulty configuration this could render all nodes inoperable at roughly the same time. There is no rollback mechanism in place that reverts faulty configuration or a gradual rollout of configuration.

I propose a 2-step approach:
1. Combine the availability checking with the configuration step, if e.g. the cluster API server becomes unavailable, the network operator will rollback the configuration on the local node and will **not** try to roll it out again, unless it was changed.
2. Gradual rollout support. A central controller is needed which renders custom resources from the cluster to individual node configuration. This individual node configuration is then used by the network operator on each node. If the rollout fails it will mark the resource as failed in the `.status` part.
The central controller renders each node configuration individually (and in a gradual fashion) thus allowing to respond to failed node configurations by stopping any further rollout until the input resources are changed again.

Because connection to the API server might be impacted by a faulty rollout, step two depends on step one. Each network-operator on the node should be individually capable to rollback the configuration to a previous, working state.

For step two: The controller (or K8s API server with ownership relations) should also clean up node configurations (which can be written as a dedicated CRD) when a node leaves the cluster. As we are a heavy user of Cluster API this is required to not clog the API server with unnecessary resources.

p-strusiewiczsurmacki-mobica:
I assume something like this will be needed
![network-operator-gradual-rollout-diagram](https://github.com/telekom/das-schiff-network-operator/assets/137421299/263c266d-d31a-41ba-9413-2ed764a499b8)

1. Network-operator working on control-plane node will be responsible for running reconciliation loop.
2. When new config is created, Operator sends the config to the worker node using gRPC endpoint.
3. Network operator that is running on worker node will have permanent storage (volume, directory mount, database etc.) to save last known good config in case a revert is needed. Storage should be permanent in case of pod's death amidst the process.
4. New configuration is being applied and connectivity is tested ((this is mostly already implemented in `healthcheck` package I think).
5. If connectivity test fails, node should restore previous config from the permanent storage.
6. Node reports the status to the control-plane gRPC endpoint.
7. If node reported success, next node (or multiple nodes in parallel) can be processed.
8. If node reported configuration is not working, all previously configured nodes should be reverted.
9. If node is unable to report the status within some time frame (6), control plane should tag config as invalid and revert other nodes that were already configured.

Other ideas are:
[...]

PR #110, opened 2024-03-27
Gradual rollout

p-strusiewiczsurmacki-mobica:
This PR implements gradual rollout as described in #98

There are 2 new CRDs added: 

`NodeConfig` - which represents node configuration
~~`NodeConfigProcess` - which represents the global state of the configuration (can be `provisioning` or `provisioned`). This is used to check if previous leader did not fail in the middle of configuration process. If so, backups are restored.~~

New pod added - `network-operator-configurator` - this pod (daemonset) is is responsible for fetching `vrfrouteconfigurations`, `layer2networkconfigurations` and `routingtables` and combining those into `NodeConfig` for each node.

`network-operator-worker` pod instead of fetching separate config resources, will now only fetch `NodeConfig`. After configuration is done, and connectivity is checked, it will backup the config on disk. If connectivity is lost after deploying new config - configuration will be restored using the local backup.

For each node there can be 3 NodeConfig objects created:
- `<nodename>` - current configuration
- `<nodename>-backup` - backup configuration
- `<nodename>-invalid` - last known invalid configuration

How does it work:

1. `network-operator-configurator` starts and leader election takes place.
2. Leader checks ~~`NodeConfigProcess`~~ if any config is in `invalid` or `provisioning` state to check if previous leader did not die amid the configuration process. If so, it will revert configuration for all the nodes using backup configuration.
3. When user deploys `vrfrouteconfigurations`, `layer2networkconfigurations` and /or `routingtables` object, `configurator` configurator will:
  -  combine those into separate `NodeConfig` for each node
  ~~- set `NodeConfigProcess` state to `provisioning`~~
5. Configurator checks new configs against known invalid configs. If any new config is equal to at least one known invalid config, deployment is aborted.
6. Configurator backups the current  <nodename> config as <nodename>-backup and deploys new config with status `provisioning`.
7. `network-operator-worker` fetches new config and configures node. It checks connectivity and:
  - if it is OK, it stores backup on disk, ant updates the status of the config to `provisioned`
  - if connectivity was lost, it restores the configuration from local backup and (if possible) updates the config status to `invalid`.
8. Configurator waits for the outcome of the config provisioning by checking the config status:
    - if status was set by worker to `provisioned` - it proceeds with deploying next node(s).
    - if status was set to `invalid` - it aborts the deployment and reverts changes on all the nodes that were changed in this iteration.
    - if it times out (e.g. node was unable to update the config state for some reason) - it invalidates the config and reverts the changes on all nodes.

Configurator can be set to update more than 1 node concurrently. Number of nodes for concurrent update can be set using `update-limit` configuration flag (defaults to 1).


chdxD1 (dismissed):
Also what happens when a node joins(!) or leaves the cluster? Especially when the joining or leaving happens mid of a rollout?

p-strusiewiczsurmacki-mobica:
> Also what happens when a node joins(!) or leaves the cluster? Especially when the joining or leaving happens mid of a rollout?

If Node joins the cluster it should be configured in next reconciliation loop iteration (I think, will check that to be sure). But, on node leave, config will be tagged as invalid (because of timeout) and configuration will be aborted. I'll try to fix that.

chdxD1:
> https://github.com/telekom/das-schiff-network-operator/pull/110/files#diff-a96964d7107a0881e826cd2f6ac71e901d612df880362eb04cb0b54eb609b8e5L70-L90

would be nice to watch node events in the central manager to create that nodeconfig **before** the next reconcile loop

p-strusiewiczsurmacki-mobica:
@chdxD1 
I had to make tons of changes, but I think the code should be much better now.
Each config now has owner reference set to node, so whenever a node is removed, all the `NodeConfig` objects should be removed automatically as well.
As for nodes added, I've introduced a `node_reconciler` which watches for nodes and whenever node is added it sends info to `config_manager` so it can trigger updates as soon as possible. It also tags deleted nodes as 'inactive' so those can be skipped if for example nod was deleted during the config deployment process.

api/v1alpha1/nodeconfig_types.go
```diff
@@ -0,0 +1,92 @@
...
+			Layer2:       []Layer2NetworkConfigurationSpec{},
+			RoutingTable: []RoutingTableSpec{},
+		},
+		Status: NodeConfigStatus{
+			ConfigStatus: "",
+		},
+	}
+}
+
+func CopyNodeConfig(src, dst *NodeConfig, name string) {
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Removed those.

cmd/operator/main.go
```diff
@@ -0,0 +1,207 @@
...
+		return fmt.Errorf("unable to create Layer2NetworkConfiguration controller: %w", err)
+	}
+
+	if err = (&controllers.RoutingTableReconciler{
+		Client:     mgr.GetClient(),
+		Scheme:     mgr.GetScheme(),
+		Reconciler: cr,
+	}).SetupWithManager(mgr); err != nil {
+		return fmt.Errorf("unable to create RoutingTable controller: %w", err)
+	}
```
schrej, line 143: I think we can just combine all of these into a single controller. That way we also avoid duplicate reconciliations in case multiple kinds of resources are updated at the same time.

cmd/operator/main.go
```diff
@@ -0,0 +1,207 @@
...
+
+	if err = (&controllers.NetworkConfigRevisionReconciler{
+		Client:     mgr.GetClient(),
+		Scheme:     mgr.GetScheme(),
+		Reconciler: ncr,
+	}).SetupWithManager(mgr); err != nil {
+		return fmt.Errorf("unable to create RoutingTable controller: %w", err)
+	}
```
schrej, line 159: Can also be combined into a single controller. Should be `For()` revisions, `Watches()` Nodes and `Owns()` NodeNetworkConfig so it als gets triggered by updates to those. Necessary to implement proper reconciliation as described below.

schrej (changes requested):
Good work overall. I'd like to change a few things though, especially with regards to the gradual rollout mechanism.

pkg/reconciler/nodeconfig_reconciler.go
```diff
@@ -0,0 +1,328 @@
...
+	timeoutCtx, cancel := context.WithTimeout(ctx, ncr.timeout)
+	defer cancel()
+
+	// wait for agent to set status to 'provisioning'
+	if err := ncr.waitForConfig(timeoutCtx, newConfig, StatusProvisioning); err != nil {
+		return fmt.Errorf("error waiting for config: %w", err)
+	}
+
+	// wait for agent to set status to 'provisioned'
+	if err := ncr.waitForConfig(timeoutCtx, newConfig, StatusProvisioned); err != nil {
```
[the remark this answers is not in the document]
schrej: Would also be great to get updates to the Revisions status with regards to the progress of the rollout.
- how many completed
- how many currently ongoing (usually one I guess, but maybe we add config for that)
- how many missing (even though it's trivial, helps looking at it at a glance)
That part of status could also be added as a printercolumn (see `+kubebuilder:printercolumn` here: https://book.kubebuilder.io/reference/markers/crd)

cmd/operator/main.go
p-strusiewiczsurmacki-mobica: Done :)

cmd/operator/main.go
p-strusiewiczsurmacki-mobica: Done. :)

pkg/reconciler/nodeconfig_reconciler.go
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Removed active waites - now each resource will be reconciled in next reconciliation loop iteration.

Also added some basic info like:
```
NAME              INVALID   ONGOING   UP-TO-DATE   QUEUED   AVAILABLE   AGE
d9de01ee2d     false                0                  4                      0                 4             10m
```

pkg/reconciler/nodeconfig_reconciler.go
```diff
@@ -0,0 +1,328 @@
...
+				return fmt.Errorf("node %s config error: %w", config.Name, InvalidConfigError)
+			}
+
+			node := &corev1.Node{}
+			if err := ncr.client.Get(ctx, types.NamespacedName{Name: config.Name}, node); apierrors.IsNotFound(err) {
+				// node was deleted - skip
+				return nil
+			}
+		}
+	}
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Removed this. :)

pkg/reconciler/config_reconciler.go
```diff
@@ -0,0 +1,182 @@
...
+	if err := r.client.Create(ctx, revision); err != nil {
+		if !apierrors.IsAlreadyExists(err) {
+			return fmt.Errorf("error creating NodeConfigRevision: %w", err)
+		}
+		if err := r.client.Delete(ctx, revision); err != nil {
+			return fmt.Errorf("error creating deleting already existing NodeConfigRevision: %w", err)
+		}
+		if err := r.client.Create(ctx, revision); err != nil {
+			return fmt.Errorf("error creating NodeConfigRevision: %w", err)
+		}
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Yes, but I've changed that to use Update instead of deletion.

pkg/reconciler/config_reconciler.go
```diff
@@ -0,0 +1,182 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	DefaultTimeout         = "60s"
+	DefaultNodeUpdateLimit = 1
+)
+
+// ConfigReconciler is responsible for creating NodeConfig objects.
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Fixed

pkg/reconciler/nodeconfig_reconciler.go
```diff
@@ -61,95 +68,208 @@ func (ncr *NodeConfigReconciler) reconcileDebounced(ctx context.Context) error {
...
-	if err := ncr.deployNodeConfigs(ctx, nodes, revisionToDeploy); err != nil {
-		return fmt.Errorf("error deploying node configurations: %w", err)
+	nodeConfigs, err := ncr.listConfigs(ctx)
+	if err != nil {
+		return fmt.Errorf("error listing configs: %w", err)
+	}
+
+	available := len(nodes)
+	ready := 0
+	ongoing := 0
```
schrej, line 90: can we return these instead of passing references?

pkg/reconciler/nodeconfig_reconciler.go
```diff
@@ -61,95 +68,208 @@ func (ncr *NodeConfigReconciler) reconcileDebounced(ctx context.Context) error {
...
-	if err != nil {
-		return fmt.Errorf("error listing revisions: %w", err)
+func getFirstValidRevision(revisions []v1alpha1.NetworkConfigRevision) *v1alpha1.NetworkConfigRevision {
+	var result *v1alpha1.NetworkConfigRevision
+	for i := range revisions {
+		if revisions[i].Status.IsInvalid {
+			continue
+		}
+		result = &revisions[i]
+		break
```
schrej, line 126: I think `return &revision[i]` would be easier to read here.

pkg/reconciler/nodeconfig_reconciler.go
```diff
@@ -61,95 +68,208 @@ func (ncr *NodeConfigReconciler) reconcileDebounced(ctx context.Context) error {
...
-			if err := ncr.client.Delete(ctx, &revisions.Items[i]); err != nil {
-				if !apierrors.IsNotFound(err) {
-					return fmt.Errorf("error deleting revision %s: %w", revisions.Items[i].Name, err)
+func (ncr *NodeConfigReconciler) processConfigs(ctx context.Context, configs []v1alpha1.NodeNetworkConfig, revision *v1alpha1.NetworkConfigRevision, ready, ongoing *int) (bool, error) {
+	for i := range configs {
+		// Every NodeNetworkConfig obejct should have 2 owner references - for NodeConfigRevision and for the Node. If there is only one owner reference,
+		// it means that either node or revision were deleted, so the config itself can be deleted as well.
+		if len(configs[i].ObjectMeta.OwnerReferences) < numOfRefs {
+			if err := ncr.client.Delete(ctx, &configs[i]); err != nil {
+				return true, fmt.Errorf("error deleting redundant node config - %s: %w", configs[i].Name, err)
```
schrej, line 137: usually a good idea to ignore not found errors when deleting. Can happen here as a race condition

api/v1alpha1/nodenetworkconfig_types.go
```diff
@@ -32,7 +32,8 @@ type NodeNetworkConfigSpec struct {
 
 // NodeNetworkConfigStatus defines the observed state of NodeConfig.
 type NodeNetworkConfigStatus struct {
-	ConfigStatus string `json:"configStatus"`
+	ConfigStatus     string      `json:"configStatus"`
+	LastModification metav1.Time `json:"lastModification"`
```
schrej, line 36: Can we call this `lastApplied`? Or what exactly is it supposed to express? Some Comments on the types would be good as they also appear in the CRD.

api/v1alpha1/networkconfigrevision_types.go
```diff
@@ -33,15 +33,23 @@ type NetworkConfigRevisionSpec struct {
 
 type NetworkConfigRevisionStatus struct {
 	IsInvalid bool `json:"isInvalid"`
+	Ready     int  `json:"ready"`
+	Ongoing   int  `json:"ongoing"`
+	Queued    int  `json:"queued"`
+	Available int  `json:"available"`
```
schrej, line 39: Maybe `Total` instead?

pkg/reconciler/nodeconfig_reconciler.go
```diff
@@ -27,14 +28,18 @@ const (
 	StatusProvisioned  = "provisioned"
 
 	controlPlaneLabel = "node-role.kubernetes.io/control-plane"
+	numOfRefs         = 2
+	configThreshold   = time.Minute * 2
```
schrej, line 32: `configTimeout` is more accurate for this one imo

pkg/reconciler/nodeconfig_reconciler.go
```diff
@@ -0,0 +1,413 @@
...
+	if err := controllerutil.SetOwnerReference(revision, c, ncr.scheme); err != nil {
+		return nil, fmt.Errorf("error setting owner references (revision): %w", err)
+	}
+
+	// prepare Layer2NetworkConfigurationSpec (l2Spec) for each node.
+	// Each Layer2NetworkConfigurationSpec from l2Spec has node selector,
+	// which should be used to add config to proper nodes.
+	// Each Layer2NetworkConfigurationSpec that don't match the node selector
+	// is removed.
+	for i := 0; i < len(c.Spec.Layer2); i++ {
```
schrej, line 306: you could use [`slices.DeleteFunc`](https://pkg.go.dev/slices#DeleteFunc) instead, probably a bit more readable.

pkg/reconciler/nodeconfig_reconciler.go
```diff
@@ -0,0 +1,413 @@
...
+	return selector, nil
+}
+
+func (ncr *NodeConfigReconciler) deployConfig(ctx context.Context, newConfig, currentConfig *v1alpha1.NodeNetworkConfig, node *corev1.Node) error {
+	var cfg *v1alpha1.NodeNetworkConfig
+	if currentConfig != nil {
+		cfg = currentConfig
+		// there already is config for node - update
+		cfg.Spec = *newConfig.Spec.DeepCopy()
+		cfg.Status = *newConfig.Status.DeepCopy()
```
schrej, line 363: do we need to deep copy here? Also, status should probably be cleared instead of copied here. Updating it is the task of the agent.

api/v1alpha1/networkconfigrevision_types.go
```diff
@@ -33,15 +33,23 @@ type NetworkConfigRevisionSpec struct {
...
 }
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:resource:shortName=ncr,scope=Cluster
 //+kubebuilder:printcolumn:name="Invalid",type=string,JSONPath=`.status.isInvalid`
+//+kubebuilder:printcolumn:name="Ongoing",type="integer",JSONPath=".status.ongoing"
+//+kubebuilder:printcolumn:name="Up-to-date",type="integer",JSONPath=".status.ready"
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Done

cmd/operator/main.go
```diff
@@ -58,12 +58,14 @@ func init() {
...
 	var configFile string
 	var timeout string
+	var updateLimit int
 	flag.StringVar(&configFile, "config", "",
 		"The controller will load its initial configuration from this file. "+
 			"Omit this flag to use the default configuration values. "+
 			"Command-line flags override configuration from this file.")
 	flag.StringVar(&timeout, "timeout", reconciler.DefaultTimeout,
 		"Timeout for Kubernetes API connections (default: 60s).")
+	flag.IntVar(&updateLimit, "update-limit", 1, "Configures how many nodes can be updated simultaneously when rolling update is performed.")
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Done

pkg/reconciler/nodeconfig_reconciler.go
p-strusiewiczsurmacki-mobica: Done

pkg/reconciler/nodeconfig_reconciler.go
p-strusiewiczsurmacki-mobica: Done

api/v1alpha1/nodenetworkconfig_types.go
p-strusiewiczsurmacki-mobica: Changed that to `lastUpdated`.
This stores the date/time when NodeNetworkConfigStatus was last updated. Is later used with `configTimeout` to invalidate config if node is unresponsive for some reason and status is not changed to `provisioned` before timeout.

pkg/reconciler/nodeconfig_reconciler.go
p-strusiewiczsurmacki-mobica: Done

pkg/reconciler/nodenetworkconfig_reconciler.go
```diff
@@ -0,0 +1,275 @@
...
+	reconciler.nodeConfig = cfg
+	if err := storeNodeConfig(cfg, reconciler.nodeConfigPath); err != nil {
+		return fmt.Errorf("error saving NodeConfig status: %w", err)
+	}
+
+	return nil
+}
+
+func (r *reconcileNodeNetworkConfig) processConfig(ctx context.Context, cfg *v1alpha1.NodeNetworkConfig) error {
+	// set config status as provisioned (valid)
```
schrej, line 142: This sets it as provisioning, not provisioned.

pkg/reconciler/nodenetworkconfig_reconciler.go
```diff
@@ -0,0 +1,275 @@
...
+	if err := setStatus(ctx, r.client, cfg, StatusProvisioned); err != nil {
+		return fmt.Errorf("error setting config status %s: %w", StatusProvisioned, err)
+	}
+
+	return nil
+}
+
+func setStatus(ctx context.Context, c client.Client, cfg *v1alpha1.NodeNetworkConfig, status string) error {
+	// set config status as provisioned (valid)
```
schrej, line 176: remove

schrej (changes requested):
Found a few more things.

One thing I dislike with the current implementation is the lack of clear responsibilities of controllers. Most importantly, only one controller should be responsible for setting the status of a resource. This is not always the case at them moment. It's perfectly fine if status takes a while to be set after a resource is created, and I'd prefer it over guessing the status when creating the resource.

api/v1alpha1/networkconfigrevision_types.go
```diff
@@ -0,0 +1,102 @@
...
+
+func NewRevision(config *NodeNetworkConfig) (*NetworkConfigRevision, error) {
+	data, err := json.Marshal(config.Spec)
+	if err != nil {
+		return nil, fmt.Errorf("error marshalling data: %w", err)
+	}
+
+	h := sha256.New()
+	if _, err := h.Write(data); err != nil {
+		return nil, fmt.Errorf("error writing MD5 data: %w", err)
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Done

api/v1alpha1/nodenetworkconfig_types.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"reflect"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NodeNetworkConfigSpec defines the desired state of NodeConfig.
+type NodeNetworkConfigSpec struct {
+	// Revision stores hash of the NodeConfigRevision that was used to create the NodeNetwokrConfig obejct.
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Done

controllers/config_controller.go
```diff
@@ -67,21 +66,12 @@ func (r *Layer2NetworkConfigurationReconciler) Reconcile(ctx context.Context, _
...
-	}
-
+func (r *ConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	err := ctrl.NewControllerManagedBy(mgr).
-		For(&networkv1alpha1.Layer2NetworkConfiguration{}).
-		Watches(&corev1.Node{}, nodesMapFn, builder.WithPredicates(nodePredicates)).
+		Named("config controller").
+		Watches(&networkv1alpha1.Layer2NetworkConfiguration{}, &handler.EnqueueRequestForObject{}).
+		Watches(&networkv1alpha1.RoutingTable{}, &handler.EnqueueRequestForObject{}).
+		Watches(&networkv1alpha1.VRFRouteConfiguration{}, &handler.EnqueueRequestForObject{}).
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: I've added a handler using `handler.EnqueueRequestsFromMapFunc`. It will always return an 1 element array of empty `ctrl.Request` so I think those should be batched by controller-runtime.

pkg/healthcheck/healthcheck.go
```diff
@@ -163,6 +152,25 @@ func (hc *HealthChecker) CheckReachability() error {
 	return nil
 }
 
+// CheckAPIServer checks if Kubernetes Api server is reachable from the pod.
+func (hc HealthChecker) CheckAPIServer(ctx context.Context) error {
+	if err := hc.client.List(ctx, &corev1.PodList{}); err != nil {
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Done

pkg/reconciler/config_reconciler.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		return fmt.Errorf("error listing revisions: %w", err)
+	}
+
+	// check if revision should be skipped (e.g. it is the same as known invalid revision, or as currently deployed revision)
+	if shouldSkip(revisions, revision) {
+		return nil
+	}
+
+	if err := cleanRevisionStatus(ctx, revision, r.client); err != nil {
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Removed that. I think that now, NodeConfig's statuses are only set by Agent, and revision statuses are only set by revision controller.

pkg/reconciler/nodeconfig_reconciler.go
```diff
@@ -0,0 +1,413 @@
...
+	StatusProvisioning = "provisioning"
+	StatusProvisioned  = "provisioned"
+
+	controlPlaneLabel = "node-role.kubernetes.io/control-plane"
+	numOfRefs         = 2
+	configTimeout     = time.Minute * 2
+)
+
+// NodeConfigReconciler is responsible for creating NodeConfig objects.
+type NodeConfigReconciler struct {
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Done

pkg/reconciler/nodeconfig_reconciler.go
```diff
@@ -0,0 +1,413 @@
...
+				// Update ongoing counter
+				ongoing++
+				// If status is 'provisioning' check for how long this status is set already, and if time threshold is exceeded, tag config as invalid.
+				if time.Now().After(configs[i].Status.LastUpdate.Add(configTimeout)) {
+					configs[i].Status.ConfigStatus = StatusInvalid
+					configs[i].Status.LastUpdate = metav1.Now()
+					if err := ncr.client.Status().Update(ctx, &configs[i]); err != nil {
+						return true, ready, ongoing, fmt.Errorf("error invalidating config: %w", err)
+					}
+					return true, ready, ongoing, nil
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: I've reworked this whole status update function. Now all revisions will have their status updated, event if the revision itself is not valid.

pkg/reconciler/nodeconfig_reconciler.go
p-strusiewiczsurmacki-mobica: Done

pkg/reconciler/nodeconfig_reconciler.go
```diff
@@ -0,0 +1,413 @@
...
+func (ncr *NodeConfigReconciler) listConfigs(ctx context.Context) (*v1alpha1.NodeNetworkConfigList, error) {
+	nodeConfigs := &v1alpha1.NodeNetworkConfigList{}
+	if err := ncr.client.List(ctx, nodeConfigs); err != nil {
+		return nil, fmt.Errorf("error listing nodeConfigs: %w", err)
+	}
+	return nodeConfigs, nil
+}
+
+func (ncr *NodeConfigReconciler) deployNodeConfig(ctx context.Context, node *corev1.Node, revision *v1alpha1.NetworkConfigRevision) error {
+	newConfig, err := ncr.createConfigForNode(node, revision)
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: You're right. Moved that down.

pkg/reconciler/nodeconfig_reconciler.go
p-strusiewiczsurmacki-mobica: You're right, there's no need for a deep copy. 
Status will only be cleared here and left to agent for further processing.

pkg/reconciler/nodenetworkconfig_reconciler.go
```diff
@@ -0,0 +1,275 @@
...
+	"github.com/telekom/das-schiff-network-operator/pkg/nl"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	defaultDebounceTime = 1 * time.Second
+
+	DefaultNodeConfigPath = "/opt/network-operator/nodeConfig.yaml"
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Changed to ` current-config.yaml`.

pkg/reconciler/config_reconciler.go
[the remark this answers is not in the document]
schrej: makes sense.

pkg/reconciler/config_reconciler.go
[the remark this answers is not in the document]
schrej: Might have sneaked back in :)

pkg/reconciler/config_reconciler.go
```diff
@@ -0,0 +1,192 @@
...
+	cr.logger.Info("new revision", "data", revision)
+
+	// get all known revisions
+	revisions, err := listRevisions(timeoutCtx, cr.client)
+	if err != nil {
+		return fmt.Errorf("error listing revisions: %w", err)
+	}
+
+	// check if revision should be skipped (e.g. it is the same as known invalid revision, or as currently deployed revision)
+	if shouldSkip(revisions, revision) {
```
schrej, line 84: can we log why we've skipped a revision? Otherwise it might be hard to figure out

pkg/reconciler/configrevision_reconciler.go
```diff
@@ -0,0 +1,466 @@
...
+		if err := crr.processConfigsForRevision(ctx, nodeConfigs.Items, &revisions.Items[i], totalNodes); err != nil {
+			return fmt.Errorf("failed to process configs for revision %s: %w", revisions.Items[i].Name, err)
+		}
+	}
+
+	revisionToDeploy := getFirstValidRevision(revisions.Items)
+
+	// there is nothing to deploy - skip
+	if revisionToDeploy == nil {
+		return nil
```
schrej, line 93: Can we add a log here? This should be a state you wouldn't want to remain in.

pkg/reconciler/configrevision_reconciler.go
```diff
@@ -0,0 +1,466 @@
...
+	}
+
+	nodesToDeploy := getOutdatedNodes(nodes, nodeConfigs.Items, revisionToDeploy)
+
+	if err := crr.updateQueueCounters(ctx, revisions.Items, revisionToDeploy, len(nodesToDeploy)); err != nil {
```
schrej, line 98: Have you tested whether this works? Considering you're updating status twice this might cause issues due to writing with an outdated generation counter failing on the first reconciliation whenever counters changed.

pkg/reconciler/configrevision_reconciler.go
```diff
@@ -0,0 +1,466 @@
...
+
+func (crr *ConfigRevisionReconciler) createConfigForNode(node *corev1.Node, revision *v1alpha1.NetworkConfigRevision) (*v1alpha1.NodeNetworkConfig, error) {
+	// create new config
+	c := &v1alpha1.NodeNetworkConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: node.Name,
+		},
+	}
+
+	c.Spec = *revision.Spec.Config.DeepCopy()
```
schrej, line 325: I don't think we need to deep copy here either, do we?

pkg/reconciler/nodenetworkconfig_reconciler.go
```diff
@@ -0,0 +1,274 @@
...
+		return nil
+	}
+	if err := r.processConfig(ctx, cfg); err != nil {
+		return fmt.Errorf("error while processing config: %w", err)
+	}
+
+	// replace in-memory working config and store it on the disk
+	reconciler.nodeConfig = cfg
+	if err := storeNodeConfig(cfg, reconciler.nodeConfigPath); err != nil {
+		return fmt.Errorf("error saving NodeConfig status: %w", err)
```
schrej, line 135: if this fails, we'll never attempt to store it again, since the reconciliation is aborted in the if statement at L113. Does it make sense to just store it to disk during each reconciliation when the in-memory revision matches the reconciled revision? Otherwise we could end up with weird state in case we ever load outdated config from disk and at the same time create a new revision (admittedly rather rare, but could happen when we roll out a new version and change config at the same time). If that new revision is invalid we would then roll back to an outdated version, which is unexpected.

schrej (changes requested):
Way cleaner now, good job.
A few more minor things :)

pkg/reconciler/config_reconciler.go
p-strusiewiczsurmacki-mobica: Done

pkg/reconciler/configrevision_reconciler.go
p-strusiewiczsurmacki-mobica: Done

pkg/reconciler/configrevision_reconciler.go
p-strusiewiczsurmacki-mobica: I've tested it and didn't see any issues. However I've changed that bit so only one status update is performed now.

pkg/reconciler/configrevision_reconciler.go
p-strusiewiczsurmacki-mobica: You're right, changed that.

pkg/reconciler/nodenetworkconfig_reconciler.go
p-strusiewiczsurmacki-mobica: I've added additional config storing  inside the if you've mentioned (L113).

schrej (approved):
lgtm now, thanks!
I'll leave it for @chdxD1 for final review

pkg/reconciler/configrevision_reconciler.go
```diff
@@ -0,0 +1,468 @@
...
+		return nil
+	}
+
+	newConfig, err := crr.createConfigForNode(node, revision)
+	if err != nil {
+		return fmt.Errorf("error preparing config for node %s: %w", node.Name, err)
+	}
+
+	if err := crr.deployConfig(ctx, newConfig, currentConfig, node); err != nil {
+		if errors.Is(err, InvalidConfigError) || errors.Is(err, context.DeadlineExceeded) {
```
chdxD1, line 307: a) I think InvalidConfigError is not used anywhere
b) if the context DeadlineExceeded happens (e.g. operator can not communicate with K8s API) I would prefer a retry instead of marking the revision as invalid

chdxD1 (changes requested):
I am fine with the overall flow and how it evolved now.

I rolled it out in one of our dev clusters and it worked, after kicking it a bit.

For some reason (**not logged!**) the config revision went into invalid state. 

I would therefore ask to add some logging which would ease debugging in a lot of cases, like when marking a config as invalid, when creating new revisions, creating new nodeconfigs etc.

pkg/reconciler/configrevision_reconciler.go
p-strusiewiczsurmacki-mobica: a) deleted InvalidConfigError
b) I've added explicitly 3 retries. If it won;t be able to connect within 3 tries, reconciler error will be reported.

Also now all actions like creation/update/deletion of objects etc. should be logged.

pkg/reconciler/configrevision_reconciler.go
```diff
@@ -0,0 +1,457 @@
...
+			case StatusInvalid:
+				// Increase 'invalid' counter so we know that there the revision results in invalid configs.
+				invalid++
+			case StatusProvisioning, "":
+				// Update ongoing counter
+				ongoing++
+				if wasConfigTimeoutReached(&configs[i]) {
+					// If timout was reached revision is invalid (but still counts as ongoing).
+					invalid++
+				}
```
chdxD1, line 170: Okay, I am struggling with this a bit. 

Imagine the following scenario:
1. A new node joins the cluster (or network operator is updated etc)
2. network-operator is not running for various reasons for 2 minutes

I am thinking of a better solution to find out if a rollout has reached a timeout.

Some things to consider:
- If a node can not rollout config at all because of *other issues* the node should be marked as NotReady by the kubelet anyway, our automation will take care of removing that node alltogether (deleting the nodeconfig anyway)
- A rollout can only timeout once it was started

What would you think of a `applyStarted` field in status which is written **before** a nodeconfig is rolled out (aka an agent has received the config but has not yet configured it). This is used to also update the lastUpdated timestamp and if applyStarted is true and lastUpdated has reached the timeout of 2 minutes the config is marked as failed.

chdxD1: Actually we already have `provisioning`, maybe just use that instead of `applyStarted`

chdxD1: but wait a second, that's already the case, let me look further..

p-strusiewiczsurmacki-mobica: What if, before invalidating the revision, we would check if node that 'timed out' is still available (e.g it was not deleted or not in `NotReady` state)? And if it was deleted or NotReady, we could then simply not use it's config's state to determine if revision should be invalidated.

p-strusiewiczsurmacki-mobica: >     * A rollout can only timeout once it was started

Currently that's true, as only the agent that operates on the node updates the `LastUpdate` timestamp in the config's status, so if there's any value there, it means that node stared to reconfigure itself.

chdxD1: I see one issue: The ConfigStatus is by default an empty string and defaults to "", the lastUpdate is 0 (aka 1990...) if we check for "" as well in line 164: `case StatusProvisioning, "":` and then check for the timeout it will always be true I guess?

ConfigStatus == "" should be a different case with a higher timeout first of all **plus** a check that lastUpdate is not 0

p-strusiewiczsurmacki-mobica: Also, made timeout value configurable using flags.

Issue #99, opened 2024-02-20
Support for NETCONF


chdxD1:
We are looking into becoming more flexible with regards to the **routing daemon** (currently FRR) running on the nodes.

A lot of routing daemons support the NETCONF/YANG configuration interface. 

As of today the configuration interface is FRR config and netlink, a pluggable implementation would be desired that would also support NETCONF/YANG.

PR #112, opened 2024-04-16
Support for plugable reconcilers

p-strusiewiczsurmacki-mobica:
This PR introduces basic support for plugable reconcilers as it will be required by #99.

`LegacyReconciler` that uses netlink/FRR was separated from the `Reconciler` struct and is provided to this struct by `IReconciler` interface, via dependency injection. Additionally stub for NETCONF based reconciler was created and config flag for selection was added.

### DIFF
--- a/pkg/reconciler/reconciler.go
+++ b/pkg/reconciler/reconciler.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	networkv1alpha1 "github.com/telekom/das-schiff-network-operator/api/v1alpha1"
 	"github.com/telekom/das-schiff-network-operator/pkg/anycast"
 	"github.com/telekom/das-schiff-network-operator/pkg/config"
 	"github.com/telekom/das-schiff-network-operator/pkg/debounce"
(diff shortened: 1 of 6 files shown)


pkg/reconciler/reconciler.go
```diff
@@ -3,92 +3,49 @@ package reconciler
...
+	getConfig() *config.Config
 }
 
-type reconcile struct {
-	*Reconciler
-	logr.Logger
+type Reconciler struct {
+	client      client.Client
+	logger      logr.Logger
+	iReconciler IReconciler
```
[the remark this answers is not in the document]
p-strusiewiczsurmacki-mobica: Done

pkg/reconciler/reconciler.go
```diff
@@ -18,32 +19,39 @@ import (
 
 const defaultDebounceTime = 20 * time.Second
 
+type Adapter interface {
```
schrej, line 22: If keep Adapter as the name here, implementations of it should also be called Adapter, not Reconciler.

schrej (changes requested):

pkg/reconciler/reconciler.go
p-strusiewiczsurmacki-mobica: Sorry, fixed that.

p-strusiewiczsurmacki-mobica:
Closing this as it will be addressed in #124

PR #138, opened 2024-08-13
Make MTU optional for VRFrouteconfiguration

watzkuh:


schrej:
Is this safe to just omit?

Took a brief look at the code, and at least in one location defaulting doesn't seem to be properly implemented: https://github.com/telekom/das-schiff-network-operator/blob/main/pkg/nl/layer3.go#L191

watzkuh:
I'm not entirely sure. I'm only assuming it because from what I can see we have the value is only set for 3 of our own VRFRouteConfigurations and unset for 816 of them* and I was hoping they weren't all broken.

(*assuming I greped correctly)

chdxD1:
@schrej good catch, that line should also contain `info.linkMTU()` however it had no impact so far

chdxD1:
@watzkuh If you could change the line mentioned by Jakob from `info.MTU` to `info.linkMTU` I approve this PR

watzkuh:
@chdxD1 Done!

chdxD1 (approved):
lgtm

PR #149, opened 2024-11-22
Set accept_untracked_na to 2 for IPv6 NA in EVPN

chdxD1:
This only works on newer Kernel versions e.g. with Ubuntu 24.04.

Older Kernel versions are out of luck.


Cellebyte:
What happens if it is not supported by the interface?

Cellebyte (approved):
LGTM 👍

PR #155, opened 2025-02-18
Converged nwop Netplan dbus client tests

p-strusiewiczsurmacki-mobica:
This PR introduces basic tests for Netplan dbus client.

Only functions that are currently used are being tested.


p-strusiewiczsurmacki-mobica:
Decided to ditch the mocking idea.

PR #156, opened 2025-02-20
FRR Config fixes

chdxD1:
- In some cases the reload does not apply all config (not even visible in show running-config). FRR might receive the netlink message of a new VRF at the same time as the config from frr-reload.py, leading to a race condition.
- Fix CommunityDrop was set to late to config
- Remove all `vrf default` parts from `router bgp <asn>` statements because the FRR behaviour changed


Cellebyte (approved):
LGTM.

It is time for a proper northbound interface.

PR #168, opened 2025-07-23
TaaS VRFs are a special VRF that should not be checked

chdxD1:


Cellebyte:
Was this not already fixed sometime in the past?

chdxD1:
> Was this not already fixed sometime in the past?

Not in this particular case, it was introduced with the check that the FRR VRFs should be configured before continuing, otherwise it tries another reload.

Issue #167, opened 2025-07-17
Allow binding metrics and webhook port to a specific address


Cellebyte:
## Context

Currently the operator configures the metrics, webhook and healthProbe binding address through a configmap.

```yaml
data:
  controller_manager_config.yaml: |
    apiVersion: controller-runtime.sigs.k8s.io/v1alpha1
    kind: ControllerManagerConfig
    health:
      healthProbeBindAddress: :7081
    metrics:
      bindAddress: :7080
    pprof:
      bindAddress: "127.0.0.1:7083"
    webhook:
      # host: "::"
      port: 7443
    leaderElection:
      leaderElect: false
```

All ports which not specifically set a `host` or `address` in front of the port will listen on `::` or in legacy on `0.0.0.0`.

## Improvement

Allow these to also be configured using the cli or via environment variables which overwrites the configmap default.
This allows a user to use the `fieldRef` to bind to `status.hostIP` when running with `hostNetwork: true`.
It also secures the specified ports from being exposed on other configured IP addresses on the host.

```yaml
## example daemonset patch to listen on alternative addresses
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: <daemon>
spec:
  template:
    spec:
      containers:
        - args:
            - --metrics-addr
            - "[$(KUBERNETES_NODE_IP)]:7080"
            - --health-addr
            - "[$(KUBERNETES_NODE_IP)]:7081"
          name: coild
          env:
            - name: KUBERNETES_NODE_IP
              valueFrom:
                fieldRef:
                  fieldPath: status.hostIP
            - name: NWOP_WEBHOOK_HOST
              valueFrom:
                fieldRef:
                  fieldPath: status.hostIP
```

Cellebyte:
Done

PR #169, opened 2025-07-24
Move bindAddresses to arguments

chdxD1:
implements #167


chdxD1:
> Can we add the ports into the manifests, to at least document them?
> 
> And after that reference them by name in the liveness and readiness probes.

Makes sense.

Cellebyte (approved):
LGTM 👍

PR #171, opened 2025-08-06
Allow NWOP to disable bridge netfilter on the fly

chdxD1:
In our deployment (with Calico or any other recent CNI) using br_netfilter is no longer required. If it is, it should be enabled per bridge (nf_call_iptables) through netlink.


Cellebyte (approved):
LGTM, but should be documented somewhere.

PR #185, opened 2025-11-07
Add SBR to VRF feature

chdxD1:
This allows to route return traffic directly to a VRF instead of consulting the cluster VRF routing table

image


Cellebyte (commented):
LGTM :)

I am unsure how a real example would look like.

Cellebyte (dismissed):
Approved 👍

PR #188, opened 2025-11-07
Fix duplicate update of status by two controllers

chdxD1:
Previously the central controller tried to erase the status while the node controller was already overwriting it.

Now the central controller will not update the status anymore. The node controller will keep track of the last rolled out revision by looking at a new status field called lastAppliedRevision.


pkg/reconciler/agent-cra-frr/nodenetworkconfig_reconciler.go
schrej: This comment is very confusing. what's actually happening here is that the status will be set to 'provisioned' unless it's already set to that.

schrej (approved):

PR #207, opened 2026-01-09
Deduplicate NodeNetworkConfig handling between CRA agents

chdxD1:
### Summary
Extracted common reconciliation logic from `agent-cra-frr` and `agent-cra-vsr` NodeNetworkConfig reconcilers into a shared `pkg/reconciler/common` package.

### Changes
- **New `pkg/reconciler/common/nodenetworkconfig.go`**: Contains shared reconciliation logic including:
  - `ConfigApplier` interface for agent-specific config application
  - Health checking and taint removal
  - Status management (provisioning/provisioned/invalid)
  - Config persistence and restoration on failure
  - Rollout management

- **Refactored `agent-cra-frr`**: Now implements `ConfigApplier` with FRR-specific logic (netlink conversion, FRR templating)

- **Refactored `agent-cra-vsr`**: Now implements `ConfigApplier` with VSR-specific logic

### Benefits
- Single source of truth for reconciliation flow
- Bug fixes only need to be made in one place
- New agents only need to implement the `ConfigApplier` interface
- Reduced code duplication (~400 lines → ~180 lines in agent reconcilers)


chdxD1:
@dehml could you kindly add your opinion here because I am touching the agent-cra-vsr code.

pkg/reconciler/common/nodenetworkconfig.go
```diff
@@ -0,0 +1,337 @@
...
+	// set NodeNetworkConfig status as provisioning
+	if err := SetStatus(ctx, r.client, cfg, operator.StatusProvisioning, r.logger); err != nil {
+		return fmt.Errorf("error setting NodeNetworkConfig status %s: %w", operator.StatusProvisioning, err)
+	}
+
+	// reconcile NodeNetworkConfig
+	if err := r.doReconciliation(ctx, cfg); err != nil {
+		// if reconciliation failed set NodeNetworkConfig's status as invalid
+		// no need to restore the config, the new one has not been applied
+		if err := r.invalidateNodeNetworkConfig(ctx, cfg, "reconciliation failed"); err != nil {
```
dehml: This is good for vsr agent because it does not need to restore previous config in case doReconciliation() failed (because invalid configs cannot be commited in netconf server), however frr agent was needed it and didn't do it now. I think that invalid configs will not be reverted anymore on frr agents.
See: https://github.com/telekom/das-schiff-network-operator/blob/main/pkg/reconciler/agent-cra-frr/nodenetworkconfig_reconciler.go#L151

dehml: That is the only difference I remember between the two reconcilers, otherwise it seems good.

chdxD1: right, good catch, thanks!

PR #289, opened 2026-05-12
fix: set source-interface to cluster VRF for IPv6 SBR in mkRule

chdxD1:


chdxD1:
@dehml could you have a look? We've seen circular routing when not setting the interface to the cluster VRF.

> Root cause: Linux kernel's IPv6 receive path sets flowi6_iif to the VRF device (cluster)

--> in general the next lookup then in the respective table does the same lookup again -> circular.
Manually fixing the source-interface in the running config solves this.

dehml:
Hi @chdxD1, if I understand it well, in case of IPv4 we have inbound interface set to TrunkInterfaceName whereas with IPv6 we have it set to ClusterVRF.Name, right ?
If that is the case I think it is cleaner to set the right value of rule.Match.Interface into the caller, the function setupPolicyRoute(). I should have done it this way from the beggining, this is not the job of mkRule to override provided rules, my bad.

chdxD1:
> Hi @chdxD1, if I understand it well, in case of IPv4 we have inbound interface set to TrunkInterfaceName whereas with IPv6 we have it set to ClusterVRF.Name, right ? If that is the case I think it is cleaner to set the right value of rule.Match.Interface into the caller, the function setupPolicyRoute(). I should have done it this way from the beggining, this is not the job of mkRule to override provided rules, my bad.

@dehml I moved it to `setupPolicyRoute` now. And exactly right, yes!

pkg/cra-vsr/layerbgp.go
dehml: You could also add the check about the SourceIP content, like you have done in setupPolicyRoute(): `if rule.Match.SourceIP != nil && *rule.Match.SourceIP != "" {`

Or create a new function that takes a TrafficMatch and returns its IP version (something like that), to be sure it will behave the same in both places.

PR #319, opened 2026-07-02
feat(sync): patch remote objects via CAPI patch helper to end Flux label war

chdxD1:
## Problem

The `network-sync` operator and a workload-cluster Flux fight over labels on every synced intent object. Two root causes:

1. **Full `Update` clobbers Flux's labels.** `applyRemote` did a read-modify-write `client.Update`, which replaces the *entire* remote object and destroys any label Flux set on it.
2. **Flux inventory labels are propagated.** `buildRemoteObject` copied *all* source labels — including `kustomize.toolkit.fluxcd.io/*` — onto remote objects, so a Flux running on the workload cluster adopts/prunes our synced objects.

Together these cause the two controllers to flap forever.

## Change

**1. Patch instead of Update — via the Cluster API patch helper.**
`applyRemote` now `Get`s the remote object (Create if absent), snapshots it with `patch.NewHelper`, mutates the *fetched* object in place, and issues a minimal `before→after` merge patch:
- `overlayBody` copies the source-of-truth payload (Spec for CRDs, Data/StringData for Secrets) without touching server-managed metadata.
- `mergeManagedMetadata` additively applies our managed-by label + source-namespace annotation, never removing foreign keys.

Because foreign labels are part of the `before` snapshot and are never modified, they never appear in the emitted patch — the sync operator stops overwriting Flux.

**2. Stop propagating GitOps metadata.**
`stripGitOpsKeys` removes `kustomize.toolkit.fluxcd.io/`, `helm.toolkit.fluxcd.io/`, and `reconcile.fluxcd.io/` prefixed labels/annotations before building the remote object, while legitimate source labels still flow through.

## Dependency

Adds `sigs.k8s.io/cluster-api v1.13.3` — the version matching this repo's pins (`controller-runtime v0.23.3` exact; `k8s.io/*` patch bump `0.35.3 → 0.35.4`). Single direct `require`.

## Tests

- `TestSyncPreservesForeignLabelsOnRemote` — a `kustomize.toolkit.fluxcd.io/name` label on the remote survives a spec-drift-correcting sync.
- `TestSyncDoesNotPropagateFluxLabels` — Flux inventory labels are stripped; legitimate labels + managed-by are kept.

## Verification

- `go build` (GOOS=linux): ok
- `go vet` / `gofmt`: clean
- `golangci-lint` (repo `.golangci.yml`): 0 issues
- Tests: 13/13 pass in `controllers/sync`

## Note

The CAPI patch helper is a merge-patch diff, not field-manager SSA (`client.Apply` + `FieldOwner`); it resolves the label war equally well. Swapping to true SSA is a small follow-up if preferred.


MaxRink (commented):
Kurzreview, kein thorough review.

Status: CI ist gruen, aber GitHub meldet aktuell `mergeStateStatus=DIRTY`; der PR ist also wegen Konflikten nicht direkt mergebar. Review ist ebenfalls noch required.

Auffaelligkeit: `controllers/sync/sync_controller.go` / `mergeManagedMetadata` ist bewusst additiv. Das schuetzt fremde Labels/Annotations, laesst aber entfernte Source-Keys und bereits frueher propagierte Flux/GitOps-Keys auf vorhandenen Remote-Objekten dauerhaft liegen. Fuer neu erzeugte Objekte ist der Ansatz plausibel; fuer bestehende Objekte kann das aber bedeuten, dass der PR den alten Label-War nicht vollstaendig aufraeumt, solange kein separater Cleanup/Recreate passiert.

Vor Merge aus meiner Sicht: Konflikte loesen und einmal bewusst entscheiden, ob diese Altlasten-Konvergenz akzeptiert ist oder aktiv bereinigt werden soll.

chdxD1:
@MaxRink thanks for the review — both points addressed.

**Merge conflicts:** Resolved by merging `main`. The only conflict was in `go.mod`/`go.sum` (`golang.org/x/oauth2`): main's dependabot bump landed `v0.33.0`, this branch needs `v0.36.0` transitively via Cluster API, so I kept `v0.36.0` and re-ran `go mod tidy` (no further changes). `golang.org/x/net` auto-merged to `v0.55.0` on both sides. Branch now merges cleanly.

**`mergeManagedMetadata` being purely additive:** You're right — additive-only protected foreign keys but left removed source keys and previously-propagated Flux/GitOps keys on existing remote objects forever, so existing objects never converged. Rather than accept the legacy convergence, I made it actively clean up.

The fix tracks the label/annotation keys *we* own in two tracking annotations (`network-sync.telekom.com/managed-labels` and `.../managed-annotations`) — the same last-applied ownership trick kubectl / server-side apply use. On each sync we now:
- prune keys we managed on the previous sync but no longer set (removed source keys, and Flux/GitOps keys once they leave our managed set), and
- still never touch foreign keys we never propagated ourselves.

This is deliberately ownership-scoped rather than a blanket strip of Flux-prefixed keys on the remote, so it keeps the truce intact: `TestSyncPreservesForeignLabelsOnRemote` (a workload-cluster Flux label we never set) still passes untouched.

One honest limitation worth calling out: for objects created by the *old* full-`Update` code (before any tracking annotation existed), a Flux key we propagated back then is indistinguishable at the object level from a legitimate workload-Flux label, so it isn't retroactively removed on the first sync — it's left alone. Everything from this change forward converges precisely. If we want to force-clean those pre-existing leftovers too, that'd need a one-off migration/recreate, which I'd rather not fold into this PR.

New regression tests: `TestSyncPrunesRemovedSourceLabel`, `TestSyncPrunesPreviouslyPropagatedFluxLabel`, `TestSyncRecordsManagedKeysOnCreate`. Full `controllers/sync` suite is green.

chdxD1:
Follow-up: rather than leaving the pre-existing Flux/GitOps leftovers in place, I now strip them outright too (commit d116599e).

Rationale: our synced objects are not part of any workload-cluster Flux inventory, so a `kustomize.toolkit.fluxcd.io/*` (or `helm.toolkit.fluxcd.io/*`, `reconcile.fluxcd.io/*`) key has no business on them regardless of who put it there — the old full-`Update` code or a controller on the workload side. If a live Flux genuinely owns the object it'll just reapply the label. The patch helper still only emits the keys we actually change, so unrelated foreign (non-GitOps) metadata is never touched.

So the metadata reconciliation now does three things per sync:
1. assert the keys we own (managed-by, source-namespace, propagated source labels);
2. prune keys we managed on a previous sync but no longer set (removed source keys) via the managed-key tracking annotations;
3. drop any GitOps/Flux-prefixed key outright.

Tests updated accordingly: `TestSyncPreservesForeignLabelsOnRemote` now asserts a foreign *non-GitOps* label survives while a Flux inventory label present on the remote is stripped; `TestSyncPrunesRemovedSourceLabel` uses a non-GitOps foreign label as the survivor. Full `controllers/sync` suite is green and `golangci-lint` passes clean.

PR #321, opened 2026-07-03
feat(l2a): surface effective interface name in Layer2Attachment status

chdxD1:
## Summary

When a `Layer2Attachment` (l2a) does **not** set `spec.interfaceName`, the `agent-hbn-l2` netplan interface defaults to `vlan.<vlan>`, where the VLAN ID comes from the `Network` referenced via `networkRef` (`Network.spec.vlan`).

Until now that effective name was only materialized downstream in the `NodeNetplanConfig` (`spec.desiredState.network.vlans`) and on the node itself. It was **not** visible on the l2a: the `InterfaceName` print column read `.spec.interfaceName`, which is blank whenever the default `vlan.<vlan>` name is used.

This PR surfaces the resolved name on the l2a so `kubectl get l2a` always shows the interface the agent creates.

## Changes

- **api**: new `Layer2AttachmentStatus.EffectiveInterfaceName`; `InterfaceName` print column repointed from `.spec.interfaceName` → `.status.effectiveInterfaceName`.
- **status**: `effectiveInterfaceName()` helper — returns `spec.interfaceName` when set, otherwise `vlan.<vlan>` derived from the referenced `Network` (empty when neither can be resolved). Mirrors the naming in `intent.buildNetplanState`, so status reflects exactly what lands on the node. Populated every reconcile in `updateLayer2AttachmentConditions`.
- **CRD**: regenerated manifest (schema field + print column).
- **tests**: `TestEffectiveInterfaceName` unit cases (default, override, empty-pointer fallback, unresolved ref, no-VLAN, override-wins-when-unresolved) + e2e assertion that the default path writes `status.effectiveInterfaceName = vlan.<vlan from networkRef>`.

## Behavior

For a Network with `vlan: 1000` and an l2a with no `interfaceName`:

```
$ kubectl get l2a
NAME     NETWORKREF   INTERFACENAME   MTU    ...
l2a-be   net-be       vlan.1000       1500   ...
```

Setting `spec.interfaceName: chris-l2` shows `chris-l2` instead.

`EffectiveInterfaceName` is a single value (node-independent, since it derives from the spec override or the Network VLAN — both node-independent).

## Testing

- `go test ./pkg/reconciler/intent/...` — all pass (incl. new unit + e2e).
- `go vet` clean on host and `GOOS=linux`.
- `gofmt` clean.

> Note: `./pkg/reconciler/agent-hbn-l2/...` cannot be compiled/tested on macOS (Linux-only `netlink`/`ethtool` syscall deps); it builds and vets cleanly for `GOOS=linux` as in CI. Unaffected by this change.

### DIFF
--- a/api/v1alpha1/network-connector/layer2attachment_types.go
+++ b/api/v1alpha1/network-connector/layer2attachment_types.go
@@ -113,6 +113,13 @@ type Layer2AttachmentStatus struct {
 	// ObservedGeneration is the most recent generation observed by the controller.
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 
+	// InterfaceName is the interface name the agent creates for this attachment:
+	// the spec.interfaceName override when set, otherwise the default
+	// "vlan.<vlan>" derived from the referenced Network's VLAN. Empty when no
+	// override is set and the Network (or its VLAN) cannot be resolved.
+	// +optional
+	InterfaceName string `json:"interfaceName,omitempty"`
+
 	// SRIOVVlanID is the VLAN ID assigned for SR-IOV device traffic.
 	// +optional
 	SRIOVVlanID *int32 `json:"sriovVlanID,omitempty"`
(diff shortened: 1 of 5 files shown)


MaxRink (commented):
Kurzreview, kein thorough review.

Status: CI gruen; GitHub ist formal noch BLOCKED wegen REVIEW_REQUIRED. Im Kernpfad wirkt der PR plausibel.

Ein kleiner plausibler Randfall bleibt: In `pkg/reconciler/intent/status/status.go` wird `effIfName` vor `statusUpdateWithRetry` berechnet und dann in der Retry-Closure gesetzt. Wenn die Status-Aktualisierung wegen Conflict neu fetched und die Layer2Attachment-Spec inzwischen geaendert wurde, kann die Closure einen stale effective interface name auf das frisch geholte Objekt schreiben. Das ist vermutlich ein kleines Race-Fenster und kein harter Blocker, aber der offene Copilot-Hinweis ist fachlich nachvollziehbar.

Aus meiner Sicht: nach Approval wahrscheinlich mergebar; den Retry-Randfall wuerde ich vorher bewusst akzeptieren oder kurz adressieren.

PR #345, opened 2026-07-23
feat(cra-grout): DPDK grout CRA flavour, vhost-user device plugin and KubeVirt binding plugin

chdxD1:
> [!NOTE]
> **Stacked PR.** Based on `feat/routed-cni-vhostuser-l2` (#344), not `main` — the
> grout changes extend files that only exist on that branch (`pkg/cni/vhostuser.go`,
> `pkg/workloadcni/server.go`, `pkg/cni/{agent,config,plugin}.go`, …). Review/merge
> #344 first; the diff here is only the grout-specific delta.

## What

Adds **`cra-grout`**, a third CRA datapath flavour: FRR control plane (BGP/EVPN via
`dplane_grout`) plus the open-source [grout](https://github.com/DPDK/grout) DPDK graph
router as the fast-path FIB owner. It is the open-source DPDK analogue of `cra-vsr`.

On top of the datapath itself this PR adds two things #344 left to a vendor component:

- **A vhost-user device plugin** (`pkg/deviceplugin`, `cmd/vhostuser-device-plugin`) —
  an open-source replacement for the 6WIND plugin used in #344. Resources, socket
  directories and the paths mounted into pods are all configurable, and the CNI derives
  the vhost-user socket mode from the device resource rather than from hard-coded
  knowledge of one vendor.
- **A KubeVirt vhost-user network binding plugin** (`pkg/kubevirthook`,
  `cmd/kubevirt-vhostuser-hook`) plus an e2e test that proves the VM datapath end to end
  against a real fabric.

BGP unnumbered now works upstream, so the previous numbered eBGP workaround is gone and
the flavour uses the same unnumbered fabric configuration as the other flavours. The
build requires a grout newer than v0.16.3.

## Commits

Rebased onto the current #344 head into fifteen commits, one per area, each buildable on
its own:

1. `feat(deviceplugin)` — the configurable vhost-user device plugin.
2. `feat(workloadcni,cni)` — the `grouttap` transport and the `tap_name` handoff. grout
   cannot adopt a moved-in kernel veth, so a pod attach is inverted: grout creates a
   `net_tap` in the CRA netns and the CNI moves it into the pod.
3. `feat(cra-grout)` — the grout datapath library (`RenderGrcli`, grcli `Batch`, mTLS
   `Manager`) and the `grout-cra` in-netns sidecar.
4. `feat(agent-cra-grout)` — the node agent, controller and reconciler, mirroring
   `agent-cra-vsr`.
5. `build(cra-grout)` — CRA and agent images, systemd units, agent overlay, Makefile.
6. `test(e2e)` — the `kind-node-grout` image and the `E2E_CRA_FLAVOR=grout` wiring (the
   default is still `frr`, so existing runs are unchanged).
7. `feat(kubevirt)` — the vhost-user network binding plugin sidecar.
8. `test(e2e)` — the grout vhost-user KubeVirt VM fixture, its test, and the CI job that
   runs it.
9. `ci(e2e)` — the e2e matrix, run as the seven jobs the pipeline is meant to have:
   cra-frr intent, cra-frr legacy, cra-frr routed CNI, cra-grout intent, cra-grout
   legacy, cra-grout routed CNI, and cra-grout vhost-user.
10. `fix(e2e)` — retry the nat64 default route until tayga is up.
11. `test(e2e)` — release VLAN-owning `Layer2Attachment`s between intent specs, so the
    twelve intent fixtures stop colliding over VLAN ownership.
12. `fix(e2e)` — bake the CNI plugins into the kind node images instead of downloading
    them at bring-up, where the download raced the nat64 gateway and its failure was
    discarded by a trailing `; true`.
13-15. `ci(e2e)` — debug collection that actually collects (the grout CRA runs systemd,
    so its logs are in the container's own journal, not in `nerdctl logs`), and a
    `timeout-minutes` on every long-running step: a *job* timeout cancels the job, so
    `if: failure()` collection never ran and the hardest failures produced no artifacts.

Everything the review rounds and the lab turned up is folded into the commit it belongs
to rather than appended as fixups: grout object naming and pruning, vhost-user socket
mode, device-plugin retry, CA validation, EAL heap and per-VRF FIB sizing and the
container memory ceiling, systemd ordering, DPDK tap naming and MACs, FRR/netns/start
ordering, VRF and L3VNI VXLAN creation, EVPN update-source, reverse-path filtering, the
node's routed path into the cluster VRF, replaying the node host routes across an FRR
restart, and several e2e gates that could report success while the state they guarded was
missing. The reasoning for each is in the code comments next to the constraint it
encodes.

## The KubeVirt binding plugin

KubeVirt is told about a binding named `vhostuser` in the KubeVirt CR
(`spec.configuration.network.binding`), with a `sidecarImage` and
`downwardAPI: device-info`. That registration is load-bearing twice over: KubeVirt then
emits **no** `<interface>` element of its own for that NIC — there is no tap to generate
— and it projects the Multus device-info file into the sidecar, which is where the
socket path and mode come from.

The sidecar serves `onDefineDomain` and splices an `<interface type='vhostuser'>` into
the domain, plus `<access mode='shared'/>` on `<memoryBacking>`. The shared access is not
cosmetic: without it the vhost-user socket connects, negotiates and then moves no
packets, because the DPDK side cannot map guest memory. The domain is edited by splicing
bytes rather than decoding and re-encoding it, so KubeVirt's `<qemu:commandline>`
namespace prefixes survive untouched.

This is a *binding plugin*, not the deprecated annotation-based hook sidecar
(`hooks.kubevirt.io/hookSidecars`). The newer declarative `Plugin` CR is Alpha and needs
KubeVirt v1.9+; e2e deploys v1.4.0, and its CEL hooks cannot read a file anyway, so this
case stays a sidecar hook regardless of version.

## Trunk support

#344 adds 802.1Q **trunk** ports, and this flavour implements them. A trunk renders as one
grout DPDK port with a `vlan` sub-interface per tagged member, each enslaved to its
member's bridge domain, plus the untagged native VLAN on the port itself.

Two constraints are encoded in `renderTrunkPorts` rather than left to chance. The port
MTU is sized from the **largest** member rather than from any one of them, because a
sub-interface cannot carry more than its parent and the smallest member would otherwise
clamp the whole trunk. And the members are pruned as a unit: grout refuses to delete an
interface that still has sub-interfaces (`EBUSY`), so the VLAN sub-interfaces are removed
before their port, which `StaleInterfaces` guarantees by returning them in that order.

## Not supported on this flavour

grout is a DPDK fast path. The other flavours implement several features with rules on
**kernel** interfaces inside the CRA netns; on grout those packets never reach the
kernel, so unless grout has an equivalent of its own the feature is simply unavailable.
The e2e suites covering these are skipped on this flavour (`framework.SkipIfGrout`)
rather than left failing. `pkg/cra-grout/README.md` carries the same table with the full
reasoning.

| Feature | State | Why |
| ------- | ----- | --- |
| **Traffic mirroring** (`MirrorSelector`/`MirrorTarget`, NNC `MirrorACLs`) | not supported — `mirror` and `intent`+`mirror` suites skipped | grout has no mirroring, ACL or classifier module, so there is nothing to render a `MirrorACL` into. `cra-frr` uses `tc` + GRE on kernel interfaces, `cra-vsr` uses the VSR's own mirroring. |
| **ARP/ND (neighbour) suppression** on stretched L2VNI segments | not supported — segments work, but ARP/ND floods the fabric | grout's bridge has no suppression knob and does not consume EVPN type-2 MAC/IP bindings for local reply. Correct, just noisier, and the noise grows with the number of endpoints on a segment. |
| **CRA-side IPv6 NAT** | not supported | grout's policy module implements `snat44`/`dnat44` only. Costs nothing today because egress NAT is node-side. |
| **Packet filtering / ACLs / QoS** | not supported | no firewall, policer or shaper module. |
| **More than one _routed_ workload-CNI attachment per node** | **supported** — IPv4 is routed over the IPv6 next-hop instead | The routed design gives *every* workload port the same on-link gateway pair, `169.254.1.1/32` plus `fe80::1/128`. grout keeps a single node-global IPv4 address table with no per-interface scope for link-local space, so the second port to ask for `169.254.1.1/32` was rejected with `EADDRINUSE` — while the IPv6 half is accepted on every port, since link-local IPv6 *is* scoped per interface. That capped a node at one routed attachment, and not gracefully: a rejected `grcli` line aborts the rest of the batch and invalidates the revision cluster-wide. **This flavour therefore never configures the IPv4 gateway at all.** Only `fe80::1/128` is programmed, and the CNI gives the pod an IPv4 default route with that address as its next-hop (`RTA_VIA`, the RFC 5549 forwarding case — `ip -4 route add default via inet6 fe80::1 dev net1`), which needs no IPv4 address on the CRA side. `cra-frr` and `cra-vsr` keep the IPv4 gateway; the kernel and the VSR both allow the duplicate. |
| **Restart without traffic loss** | not supported | a grout restart empties the FIB *and* the interface table — ports, uplinks and VRFs are gone and nothing rebuilds them, so the node has to re-run its CRA bootstrap. Restarting *FRR* is not a problem: verified against grout 0.16.3 with FRR 10.6.1, neither an FRR restart nor a `frr-reload.py` drops the L3VNIs or removes grout FIB entries programmed out-of-band with `grcli`, so the sidecar hot-reloads FRR on a config change. |
| **Remote VTEPs advertising only IPv6 prefixes** | fixed upstream — needs an image carrying the fix (`:edge`, not the v0.16.3 tag) | `dplane_grout` cached an EVPN next-hop's RMAC keyed on (vrf, address) *including the address type*, exactly as zebra handed it over, but normalised an IPv4-mapped gateway (`::ffff:a.b.c.d`) back to plain IPv4 before looking it up when building the L3 next-hop. A VTEP zebra only ever presents in the mapped form — what a device advertising nothing but IPv6 prefixes produces — was cached under a key the lookup never used: its next-hop stayed unresolved on the SVI instead of becoming `flags=static remote` on the VXLAN port, and its share of an ECMP pair silently blackholed. Found bringing this flavour up on 0.16.3 with FRR 10.6.1, fixed upstream, no longer reproducible on `:edge`. Listed because it is one of the two reasons this flavour tracks `:edge`. |
| **Kernel-visible datapath state** | by design | `dplane_grout` disables zebra's kernel namespace: `ip`/`tcpdump` on the CRA show nothing about forwarded traffic. Use `grcli` instead. |

One rejected address has a larger blast radius than one port: a failing `grcli` aborts
the whole batch, which marks the NodeNetworkConfig and its `NetworkConfigRevision`
invalid, and the operator then refuses to deploy *any* revision cluster-wide. Because
revision names are content hashes, regenerating the same config reproduces the same
invalid revision, so there is no self-recovery — see the README.

Beyond the datapath, grout needs hugepages, a DPDK-capable interface and a privileged
CRA, so the flavour is not a drop-in for clusters that cannot give the CRA those.

## Testing

- `GOOS=linux go build ./...`, `go vet ./...` ✅
- `go test` on Linux for `pkg/kubevirthook`, `pkg/deviceplugin`, `pkg/cra-grout`,
  `pkg/cni`, `pkg/workloadcni` ✅
- `golangci-lint` (the version CI pins): no new issues over the base branch ✅
- **Lab, `E2E_CRA_FLAVOR=grout E2E_KUBEVIRT=1 E2E_KUBEVIRT_VHOSTUSER=1`, DPDK-capable
  host: the full vhost-user KubeVirt suite passes, 6/6 specs.** Both clusters come up
  through `kubeadm`, underlay BGP and EVPN converge on every node, type-5 routes are
  imported into the cluster VRF, KubeVirt installs, the device plugin advertises and
  allocates a socket, the binding plugin converts the domain, grout creates the
  `net_vhost` port and enslaves it to the right bridge domain, and a pod on VLAN 501
  **pinned to the other node** pings the VM dual-stack with 0% loss — so the traffic
  crosses the real fabric (pod → macvlan → leaf → VXLAN → the other node's grout →
  bridge domain → vhost-user socket → guest) rather than looping back inside one node.

The suite is wired into CI as its own `e2e-kubevirt-vhostuser` job. It cannot share a lab
with the existing e2e jobs: the node image bakes in a different CRA, hugepages have to be
reserved before kubelet starts, and the fixture switches the operator to the intent
reconciler, which replaces the legacy pipeline.


chdxD1:
@copilot review

PR #348, opened 2026-07-30
feat: render listen-range BGP peers for management VRF

chdxD1:
## Problem

BGP peers (including `listen-range`) configured on the management VRF were silently dropped by both agents.

**Root cause** — two independent gaps:

### cra-frr (`config/agent-cra-frr/frr.conf.tpl`)
The `FabricVRFs` loop that calls `bgpNeighbor` explicitly skips the management VRF name. The dedicated `router bgp ... vrf <mgmt>` section looked into `FabricVRFs[mgmtName]` for VRF imports but never rendered BGP peers.

### cra-vsr (`pkg/cra-vsr/layerbgp.go`)
`setupManagementVRF()` read static routes and VRF imports from `FabricVRFs[mgmtName]` but had no `BGPPeers` loop, unlike `setupFabricVRF` / `setupLocalVRF`.

The input/builder layer (both new intent and legacy operator path) was unaffected — it placed the peer into `FabricVRFs[mgmtName]` correctly; only the agents discarded it.

## Fix

- **cra-frr**: add a conditional block inside the management VRF `router bgp` section that iterates `FabricVRFs[managementVRFName].BGPPeers` and calls `bgpNeighbor`.
- **cra-vsr**: add the missing `BGPPeers` loop inside the `conf` guard in `setupManagementVRF()`.

## Tests

New cra-vsr unit test: wires a `listen-range` peer into the management VRF's `FabricVRF` entry and asserts the generated VRouter XML contains a listen block with the expected CIDR.

### DIFF
--- a/config/agent-cra-frr/frr.conf.tpl
+++ b/config/agent-cra-frr/frr.conf.tpl
@@ -411,6 +411,12 @@ router bgp {{ $.Config.LocalASN }} vrf {{ $.Config.ManagementVRF.Name }}
   bgp router-id {{ $.Config.VTEPLoopbackIP }}
   no bgp suppress-duplicates
   no bgp default ipv4-unicast
+  {{ if containsKey $.NodeConfig.FabricVRFs $.Config.ManagementVRF.Name }}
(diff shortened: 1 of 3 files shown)


dehml:
Hi Chris, this is OK for me ;-)

PR #358, opened 2026-08-14
chore(bpf): regenerate bindings for cilium/ebpf v0.22.0

MaxRink:
## Root cause

`bpf2go` ships with `cilium/ebpf`. Its generated output changed in v0.22.0:

- it emits object name constants for maps and programs
- it uses `any` instead of `interface{}`

The repo commits the `bpf2go` output in `pkg/bpf/nwopbpf_*_bpfel.go`. The `validate generated files` job runs `make generate` and fails on any diff. Dependabot bumps the module but cannot run `make generate`, so the job fails on every `cilium/ebpf` bump.

## Change

Bump `cilium/ebpf` to v0.22.0 and commit the regenerated bindings in the same change.

## Unblocks

- #342 (`go-minor-patch` group, 7 updates). This removes the `cilium/ebpf` part of that group as a cause of the `validate generated files` failure.

## Verification

- `make generate` produces no further diff
- `go build ./...` and `go vet ./...` pass
- `go test` over the `make test` package set passes
- `golangci-lint run ./...` reports 88 issues, identical to the `main` baseline

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/telekom/das-schiff-network-operator
 go 1.25.0
 
 require (
-	github.com/cilium/ebpf v0.21.0
+	github.com/cilium/ebpf v0.22.0
 	github.com/cnf/structhash v0.0.0-20250313080605-df4c6cc74a9a
 	github.com/coreos/go-systemd/v22 v22.7.0
 	github.com/go-logr/logr v1.4.3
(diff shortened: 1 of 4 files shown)


MaxRink:
Superseded by #366 as the canonical generated-file remediation for current Dependabot PR #362. The generated BPF patches in #358 and #366 are byte-for-byte identical; #362 already contains the same cilium/ebpf v0.22.0 module bump. Closing this older standalone PR to avoid duplicate work. Related: #342, #362, #366.

PR #367, opened 2026-08-24
fix: align FRR repository with Ubuntu 26.04

MaxRink:
## Purpose

Follow-up to Dependabot PR #364. The PR moves the CRA-FRR runtime image to Ubuntu 26.04 (resolute), but the Dockerfile still selected the Ubuntu noble FRR repository and an FRR train not published for resolute.

## Changes

- use the resolute FRR repository suite;
- select FRR 10.7, which is published for Ubuntu 26.04.

## Validation

- verified the resolute/frr-10.7 repository metadata is available;
- git diff --check.

Related: #364.

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -294,10 +294,10 @@ jobs:
             --kubeconfig=/etc/kubernetes/admin.conf get events -A \
             --sort-by='.lastTimestamp' > /tmp/e2e-logs/cluster2-events.log 2>&1 || true
           docker exec nwop-control-plane kubectl \
-            --kubeconfig=/etc/kubernetes/admin.conf get nodes -o yaml \
+            --kubeconfig=/etc/kubernetes/admin.conf --request-timeout=15s get nodes -o yaml \
             > /tmp/e2e-logs/cluster1-nodes.yaml 2>&1 || true
           docker exec nwop2-worker kubectl \
-            --kubeconfig=/etc/kubernetes/admin.conf get nodes -o yaml \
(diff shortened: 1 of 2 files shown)


MaxRink:
The current E2E failure is a real FRR routing regression, not an image-build failure.

On `dbeef93a548bf55afed9ee15f0f754c142dd1921`, [run 34231377897](https://github.com/telekom/das-schiff-network-operator/actions/runs/34231377897) built `frr 10.7.1-0~ubuntu26.1`. The added diagnostics show worker IPv4 host routes present in the Linux FIB but absent from FRR's RIB/EVPN advertisements. Workers retain the forward route to the API VIP; return traffic instead matches the control plane's aggregate blackhole route, and workers become unreachable.

This matches [FRRouting/frr#23020](https://github.com/FRRouting/frr/issues/23020): an unrelated interface-down event removes IPv4 kernel routes through interfaces without IPv4 addresses from zebra. The regression remains in 10.7.1; [the proposed upstream fix](https://github.com/FRRouting/frr/pull/23234) is still open.

I independently reproduced the same IPv4-via-IPv6 `proto static` route shape in isolated official FRR containers: both versions initially learn the route; after deleting an unrelated dummy interface, 10.7.1 loses it while Linux retains it, whereas 10.6.2 retains it throughout the observation window. This is a focused reproduction, not a full patched-image E2E pass.

Keeping this PR blocked rather than masking the failure with a default-route bypass or skipped tests. The older FRR train is not published for resolute; a custom patched FRR build would be a separate implementation decision. No such dependency replacement has been made here.

PR #368, opened 2026-08-26
fix(macvlan): poll vlan.* FDB for stale MACs and clean bridge port entries

eumel8:
# Summary
Rewrite the MAC sync mechanism to fix IPv6 DAD failures during StatefulSet pod rolls on K8s 1.34+.

# Problem
When a macvlan pod is deleted during a StatefulSet rolling update, the kernel removes the self permanent FDB entry on the host-side vlan.XXXX interface. However, the bridge-learned entry on l2v.XXXX (master l2.XXXX) persists until the bridge ageing timer expires (300s). During this window, FRR continues to advertise the stale (IP, oldMAC) binding via EVPN Type-2 routes. When the replacement pod starts with the same IP but a new random MAC, IPv6 DAD fails because remote nodes still respond with the stale MAC binding.
This became visible in K8s 1.34 because the time between pod deletion and pod recreation narrowed (faster kubelet startup, kernel 6.8.0-100 bridge behaviour) below the EVPN convergence time.

# Root Cause
1. Pod deleted → macvlan interface destroyed → self permanent FDB on vlan.XXXX removed by kernel
2. Bridge-learned entry on l2v.XXXX persists (bridge ageing 300s, no kernel notification)
3. FRR/Zebra still sees MAC in bridge FDB → keeps advertising EVPN Type-2 route
4. New pod created (same IP via Whereabouts IPAM, new random MAC) → DAD runs → remote nodes respond with stale MAC → DAD fails
5. Kernel does NOT retry DAD after failure → pod stuck in dadfailed tentative until manual deletion

# Fix
Rewrite pkg/macvlan/macvlan.go:
- Poll each vlan.XXXX interface every 1 second for NTF_SELF unicast MACs
- When a MAC disappears (pod deleted), immediately delete the corresponding bridge-learned FDB entry from l2v.XXXX
- FRR/Zebra detects the removal and withdraws the EVPN Type-2 route within 1s
- Replacement pod's DAD succeeds because remote nodes no longer hold stale binding
The old approach (polling the l2.XXXX bridge for self permanent entries and mirroring to a dummy interface) never worked because macvlan MACs appear as self permanent on the physical vlan.XXXX interface, not on the bridge itself.

# Breaking Change
The semantics of --macvlan-interface-prefix changed: it now refers to the vlan.* interface prefix (previously referred to the bridge l2.* prefix). The DaemonSet args must be updated to --macvlan-interface-prefix=vlan..

# Testing
Tested on a production-like cluster with Kubernetes 1.34.12, Kernel 6.8.0-100, FRR-based EVPN, Multus macvlan + Whereabouts IPAM:
- Without fix: DAD fails consistently on StatefulSet pod rolls (macvlan interfaces stuck in dadfailed tentative)
- With fix: Stale MAC removed within 1s of pod deletion, DAD passes, replacement pod reaches Ready with all IPv6 addresses functional

### DIFF
--- a/pkg/macvlan/macvlan.go
+++ b/pkg/macvlan/macvlan.go
@@ -2,7 +2,6 @@ package macvlan
 
 import (
 	"bytes"
-	"errors"
 	"fmt"
 	"net"
 	"strings"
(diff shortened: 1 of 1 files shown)


chdxD1 (commented):
should be fine, please make the linter pass and look especially at the first copilot comment.

I will check next week how this could work on main, because responsibility between the two interfaces is split in the CRA design. One option in our case would be to drop DAD completely.

eumel8:
Closing this PR — after further investigation and testing, the FDB cleanup approach is insufficient to solve the IPv6 DAD failure problem:
1. Race condition is unwinnable: Even with 1s polling, the replacement pod's DAD completes before the stale FDB entry can be removed and the EVPN withdrawal propagated to remote nodes.
2. Remote nodes re-learn stale entries: After local deletion, FRR/Zebra on remote nodes re-programs the stale MAC via EVPN within seconds, creating an infinite delete/re-learn loop.
3. Actual fix is at a different layer: The working solution is a CNI meta-plugin (cni-nodad) that sets net.ipv6.conf.default.dad_transmits=0 inside the pod network namespace before the macvlan plugin assigns the IPv6 address. This prevents DAD entirely, making the EVPN stale state irrelevant.
The stale FDB entries self-correct via EVPN MAC-mobility sequence numbers within minutes and cause no harm when DAD is disabled.

PR #374, opened 2026-09-01
feat(workload-cni): add VSR vhost-user CNI handoff

chdxD1:
## Summary

Adds the VSR-targeted vhost-user CNI handoff on top of the routed/L2 attachment stack:

- adds vhost-user transport and socket fields to workload-port APIs and the gRPC payload;
- resolves 6WIND device-plugin allocations, publishes the pod-side device info, and sends the host-side socket to the node-local agent;
- rejects this unsupported transport on CRA-FRR and adds the Vhost KubeVirt example.

Socket mode is selected from required NAD userdata `socket_mode` (`client` or `server`). Current 6WIND device-plugin device-info always reports `server`, so its mode is deliberately ignored; device info remains the source for the allocated pod-side path.

CRA-VSR NETCONF realization is intentionally owned by #356.


pkg/workloadcni/server.go
```diff
@@ -55,6 +55,10 @@ const (
 	// it as infra-<ifname>, and that reference must itself fit the kernel limit,
 	// so the name has that much less room. Mirrors the CRD's MaxLength.
 	maxInterfaceNameLen = kernelIfNameLen - len(InfraPortPrefix)
+	// maxVhostInterfaceNameLen bounds a vhost-user CRA-side interface name. Its
+	// VSR reference is fpvhost-<ifname>, which is longer than infra-<ifname>, so
+	// the name gets correspondingly less room.
+	maxVhostInterfaceNameLen = kernelIfNameLen - len(FpvhostPortPrefix)
```
dehml: I'm not sure this is needed to take in account "fpvhost-" and "infra-" into the interface name, why do you think it is ?
These prefixes will not be present into the interface names:
- For fpvhost the prefix is used for internaly named the "port", the YANG model does not limit its size.
- For infrastructure the prefix is used for interface alias which is limited to 255 characters.

Am I wrong somewhere ?

chdxD1: right - I actually overdid a fix here.. `IFALIASZ` is 256 (255) chars, which however **only** applies to the veths (that need the `infra-` prefix in their alias), not to fpvhost interfaces.

ofc vSR interfaces (checked in the `vrouter-types.yang`) limits ifname to 15 characters (same like Linux) but we don't need the interface name to have the prefixes at all.

chdxD1: btw using the `fpvhost-` prefix for fast path interfaces was derived from the `hna-evpn` deployment guide.

```
/ system fast-path virtual-port fpvhost fpvhost-{{hna_net.name}}
/ vrf main interface fpvhost vho-{{hna_net.name}} port fpvhost-{{hna_net.name}}
```

dehml: It is required to prefix virtual-port fpvhost name with "fpvhost-" but this name is for internal usage and has no length constraint (first line).
However the linux interface name (second line) which has the length constraint is prefixed by "vho-" in the deployment-guide, but this is not mandatory, you can named it as you wish, without any prefix, and this is actually what you are doing (in addWorkloadPorts()).

PR #384, opened 2026-09-15
fix(cra-vsr): verify layer2 config actually applied after NETCONF commit

Haven456:
## Summary

Adds a post-commit verification step to the VSR NETCONF apply path so that partially-applied layer2 configuration is detected and reported instead of silently marked as successful.

## Problem

On a production cluster, a customer-facing VLAN never got its bridge/vxlan instantiated on the VSR, while sibling VLANs on the same node were fine. The NodeNetworkConfig status showed "provisioned" with no errors logged, so nothing indicated a problem until the customer reported it.

`Manager.ApplyConfiguration()` sends the full desired config as one NETCONF `edit-data`+`commit` transaction and treats an "ok" RPC reply as full success. It never reads back the running config to confirm every bridge/vxlan/vlan object was actually instantiated. The generic post-apply healthcheck can't catch this either, since it only checks a static interface list on the agent's own host netns, not the VSR's `hbn` netns.

Is this a known VSR-side issue (e.g. an internal race/ordering bug when applying multiple layer2 entries in one transaction)? Would appreciate input from anyone with more context on the VSR internals here — from the operator side we can only confirm the commit itself returned "ok" while part of the config didn't land.

## Solution

After `Commit()` succeeds, read back the `Running` datastore and verify each `Layer2` entry from the NodeNetworkConfigSpec actually landed — bridge exists (in the right VRF), vxlan and vlan interfaces exist, and both are enslaved to the bridge. If anything is missing, return an error instead of marking success.

This plugs into the existing reconciler machinery unchanged: the error causes the NodeNetworkConfig to be marked `invalid` (not `provisioned`), which the ConfigRevisionReconciler already surfaces as an invalid revision — making the failure visible instead of silent.

## Test scope

Added unit tests in `pkg/cra-vsr/verify_test.go` reproducing the incident scenario (bridge present for one VLAN, missing for a sibling), confirming it's now caught as an error. `go test ./pkg/cra-vsr/... ./pkg/reconciler/...` passes; `golangci-lint run --new-from-rev=main` reports 0 new issues.


chdxD1:
As we expect a COMMIT to work on the vSR, this should be a ticket to 6WIND with a troubleshooting-report.

